Issue #2, opened 2022-12-23
Configure app for auto deploys from GitHub


apbendi:
- [ ] Should build and deploy preview when a PR is opened
- [ ] Should build and deploy to production from the `master` branch when a PR is merged

PR #5, opened 2023-01-02
Main page scaffold

### LINKED ISSUE #4: Scaffold Out Basic Initial UI
Take an initial pass laying out the basic UI. Add components for recipient, token selection, amount, and "Shield" button. UI should approximate the [Umbra send page](https://app.umbra.cash/send). Add 1-2 tokens loaded from a hardcoded [token list](https://uniswap.org/blog/token-lists).

alexkeating:
#### Description

Added form fields for the initial UI with Chakra. I chose Chakra because its comprehensive component library will allow us to build more quickly with less upfront effort relative to styled components. Also, be aware there are a few placeholders for some things we will need in the future such as a logo and theme color scheme.

Closes #4


package.json
```diff
@@ -9,8 +9,12 @@
     "lint": "next lint"
   },
```
wildmolasses, line 10: can we pin node/yarn versions via volta? shall we try yarn 3?

yarn.lock
```diff
@@ -2,6 +2,13 @@
 # yarn lockfile v1
 
 
+"@babel/code-frame@^7.0.0":
```
wildmolasses, line 5: happy we are using yarn over npm; README still references npm, can we correct it?

pages/index.tsx
```diff
@@ -1,83 +1,88 @@
...
+import {
+  FormControl,
+  FormLabel,
+  FormErrorMessage,
+  FormHelperText,
+} from "@chakra-ui/form-control";
+import type { NextPage } from "next";
+import Head from "next/head";
+
+const exampleTokenList = [
```
wildmolasses, line 14: is there any sense in trying to componentize more of this index page? reluctant to over-engineer given that it's an SPA, curious for your thoughts @alexkeating

wildmolasses (commented):
looking fantastic!! a couple small comments; also, i'm currently getting this unhandled error on first load:

Screenshot 2023-01-03 at 10 37 58 AM

pages/index.tsx
```diff
@@ -1,83 +1,88 @@
...
+            size="lg"
+            placeholder="vitalik.eth"
+            pr="4.5rem"
+            height="4rem"
+            mb=".75rem"
+          />
+        </FormControl>
+        <FormControl>
+          <FormLabel>Token</FormLabel>
+          <Select placeholder="Ether" size="lg" height="4rem" mb=".75rem">
```
wildmolasses, line 54: not merge blocking, just a comment: there are some additions we may want to make here, for example including the token address next to the token name, perhaps including the token icon as well. We may have to reply on a different component depending on how chakra does `<Select>`

components/Header.tsx
```diff
@@ -0,0 +1,18 @@
+import { Image } from "@chakra-ui/image";
```
mds1, line 1: _(arbitrary comment location to start a thread)_

I got the same "Unhandled Runtime Error" that @wildmolasses mentioned [here](https://github.com/ScopeLift/railgun-shielder/pull/5#pullrequestreview-1234717650)

package.json
```diff
@@ -9,8 +9,12 @@
...
   "dependencies": {
+    "@chakra-ui/react": "^2.4.6",
+    "@emotion/react": "^11",
+    "@emotion/styled": "^11",
     "@rainbow-me/rainbowkit": "^0.8.1",
     "ethers": "^5.0.0",
+    "framer-motion": "^6",
     "next": "^12.1.6",
```
mds1, line 18: Should we use next.js v13? That's the default used when I scaffold a new project, with create-next-app or create-wagmi, so wondering why we're on v12

pages/_app.tsx
```diff
@@ -1,17 +1,39 @@
-import "../styles/globals.css";
 import "@rainbow-me/rainbowkit/styles.css";
```
mds1, line 1: Do we want their styles? Maybe they are just good defaults and we'll override things later?

pages/_app.tsx
```diff
@@ -1,17 +1,39 @@
...
+import { extendTheme } from "@chakra-ui/theme-utils";
 import type { AppProps } from "next/app";
+import Head from "next/head";
 import { RainbowKitProvider, getDefaultWallets } from "@rainbow-me/rainbowkit";
 import { configureChains, createClient, WagmiConfig } from "wagmi";
-import { mainnet, polygon, optimism, arbitrum, goerli } from "wagmi/chains";
+import {
+  mainnet,
+  polygon,
+  optimism,
```
mds1, line 12: Can remove this unused import

pages/index.tsx
mds1, line 14: Using a TODO makes it easier to find that this need to be changed, and minimizes that diff since var names won't need to be changed. Alternatively, track this in an issue?
```suggestion
// TODO this is an example list, update token list with real list of supported tokens
const tokenList = [
```

components/Header.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Flex } from "@chakra-ui/layout";
```
mds1, line 2: _(arbitrary comment location to start a thread)_

I'm getting tons of 429's in the console because we use a default provider that's getting rate-limited. Can we pull the RPC info out into an environment variable?

package.json
alexkeating: Yes, I can make this change

pages/index.tsx
alexkeating: Yes, definitely think we should componentize more. My goal with this pr was just to put in a scaffold, and I figured we could handle pulling out these components as we added interactive functionality as any extra work now would probably be redone.

pages/index.tsx
alexkeating: +1 Chose a select for simplicity, but we will need much more functionality. I figured we can handle that in a separate ticket/pr

package.json
alexkeating: Not sure, I used `yarn create @rainbow-me/rainbowkit@latest`. I can try upgrading to 13

pages/_app.tsx
alexkeating: Yea, they are decent defaults and we can refine later.

pages/_app.tsx
alexkeating: Yes, and we should add a prettier check and command to catch this stuff prior to review

pages/index.tsx
```diff
@@ -1,83 +1,88 @@
...
-import Head from 'next/head';
-import styles from '../styles/Home.module.css';
+import { Button } from "@chakra-ui/button";
+import { Input, InputGroup, InputRightElement } from "@chakra-ui/input";
+import { Box, Flex, Heading, Text } from "@chakra-ui/layout";
+import { Select } from "@chakra-ui/select";
+import {
+  FormControl,
+  FormLabel,
+  FormErrorMessage,
```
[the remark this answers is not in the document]
alexkeating: Ah sorry about this setup a new computer during the break and my editor is probably not setup properly

components/Header.tsx
alexkeating: We have the Alchemy provider setup. If you copy the .env.sample and fill in the API_KEY env var with an alchemy token those should go away. I can probably update the README to be clearer.

package.json
mds1: Feel free to create an issue and handle in a separate PR, small PRs are nice 🙂

pages/_app.tsx
mds1: Good call, can you create an issue to setup CI that runs prettier/linters/builds the app? We can add tests too of course once we have that. Up to you whether to adds this on as part of https://github.com/ScopeLift/railgun-shielder/issues/2 and do that early, or to handle it separately from the deployments

components/Header.tsx
mds1: Ah yea I didn't realize we needed those env vars, the names seem generic so I thought they were defaults from scaffolding. Let's rename to `NEXT_PUBLIC_ALCHEMY_API_KEY` to be more specific

And wdyt about deleting `NEXT_PUBLIC_ENABLE_TESTNETS` and instead doing `...(process.env.NODE_ENV === 'development' ? [goerli] : [])`?

apbendi: > And wdyt about deleting NEXT_PUBLIC_ENABLE_TESTNETS and instead doing ...(process.env.NODE_ENV === 'development' ? [goerli] : [])?

I don't think we should do this, actually. Supporting Goerli (on the production deploy) is one of the requirements.

components/Header.tsx
apbendi: Once I added the Alchemy API key this went away. Did you have the same experience @mds1?

components/Header.tsx
alexkeating: Removed testnet check

pages/_app.tsx
alexkeating: Created an issue #7

package.json
alexkeating: Created and issue #6

components/Header.tsx
mds1: Ah yes that fixed it, marking this resolved

components/Header.tsx
mds1: ah good call @apbendi so yea we don't any conditionals, marking this resolved

mds1 (approved):
LGTM!

wildmolasses (approved):
lgtm! thanks @alexkeating

PR #25, opened 2023-01-05
Add ci

### LINKED ISSUE #7: Setup CI that runs prettier, linter, builds and deploys the app.
"Good call, can you create an issue to setup CI that runs prettier/linters/builds the app? We can add tests too of course once we have that. Up to you whether to adds this on as part of https://github.com/ScopeLift/railgun-shielder/issues/2 and do that early, or to handle it separately from the deployments"

_Originally posted by @mds1 in https://github.com/ScopeLift/railgun-shielder/pull/5#discussion_r1060998876_

alexkeating:
#### Description

1. Add prettier and eslint
2. Add a ci to check lint, build and  format

Closes #7


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,26 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
+      - "main"
+  push:
+    branches:
+      - "main"
```
mds1, line 9: Nit, recommend using these triggers. The main difference is running on all PRs (sometimes we'll merge one branch into another), and ability to manually dispatch the workflow if needed
```suggestion
on:
  workflow_dispatch:
  pull_request:
  push:
    branches:
      - main
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
```
mds1, line 18: Our package json has node 18.12.1, these should probably match?
```suggestion
          node-version: 18.12.1
```

.prettier.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+  overrides: [
+    {
+      files: "**/*.sol",
+      options: {
+        printWidth: 120,
+        tabWidth: 2,
+      },
+    },
```
mds1, line 9: We don't have any solidity files so let's delete this

.prettier.js
```diff
@@ -0,0 +1,22 @@
...
+    {
+      files: ["**/*.js", "**/*.ts", "**/*.vue"],
```
mds1, line 11: Should have jsx/tsx instead of vue files. But maybe we just remove this key so it runs against all supported files? e.g. JSON and markdown too

.prettier.js
```diff
@@ -0,0 +1,22 @@
...
+      options: {
+        bracketSpacing: true,
+        trailingComma: "es5",
+        tabWidth: 2,
+        printWidth: 120,
```
mds1, line 16: We recently decided that 100 is what we use for solidity projects, let's match that here too (umbra will prob stay at 120 which I think is where this was from)
```suggestion
        printWidth: 100,
```

.prettier.js
```diff
@@ -0,0 +1,22 @@
...
+    {
+      files: "**/*.sol",
+      options: {
+        printWidth: 120,
+        tabWidth: 2,
+      },
+    },
+    {
+      files: ["**/*.js", "**/*.ts", "**/*.vue"],
+      options: {
```
mds1, line 12: There's a package called `@trivago/prettier-plugin-sort-imports` which extends prettier to sort imports, then I like configuring it like this:

```
// Standard prettier options
"printWidth": 100,
"tabWidth": 2,
// @trivago/prettier-plugin-sort-imports options
"importOrder": ["^react(.*)", "next/(.*)", "<THIRD_PARTY_MODULES>", "@/(.*)", "^[./]"],
"importOrderSortSpecifiers": true
```

This does two things:
1. Changes `import {b, a} from 'foof'` to be alphabetically sorted to `import {a, b} from 'foo'`
2. Sorts paths by the order specified in the array above. We'd just need to add the below snippet to tsconfig.json and then import from our own dirs using the `@/` syntax which I think is common. And `<THIRD_PARTY_MODULES>` is a specially recognized keyword by the package. 

Lmk what you guys think of this, but personally I find consistent import ordering nice, especially when it comes for ~no cost

```js
{
  "compilerOptions": {
    // --- snip ---
    "paths": {
      "@/*": ["./src/*"]
    }
  },
}

```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,26 @@
...
+          cache: "yarn"
+      - run: yarn install
+      - name: Lint
+        run: yarn run lint
+      - name: Format
+        run: yarn run checkFormat
+      - name: Build
+        run: yarn run build
```
mds1, line 26: If linting fails, we should still run the fmt/build checks. Can do this by making them separate jobs, or by adding `if: always()`. I'm ok with either
```suggestion
      - name: Format
        if: always()
        run: yarn run checkFormat
      - name: Build
        if: always()
        run: yarn run build
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,26 @@
...
+on:
+  pull_request:
+    branches:
+      - "main"
+  push:
+    branches:
+      - "main"
+
+jobs:
+  build:
```
mds1, line 12: this job contains lint, fmt, build, so just renaming the job
```suggestion
  CI:
```

mds1 (commented):
LGTM overall, lots of small comments / easy fixes

.prettier.js
alexkeating: I am down to use this

mds1: (also noting that the two items I listed under "standard prettier options" are just an example, feel free to leave all the options you already have)

alexkeating:
Made the suggested changes, hopefully it should be good to go

.prettier.js
wildmolasses: i'm cool with this.

bikeshedding a bit but what about sorting package.json via sort-package-json?

wildmolasses (approved):
lgtm

.prettier.js
mds1: Isn't package.json sorted by default when doing `yarn add <pkg>`?

wildmolasses: sort-package-json sorts the scripts, packages, other fields

mds1 (approved):

.prettier.js
mds1: ahh got it, up to you guys. those rarely change so I'm also ok with just sorting manually to avoid introducing another dependency (I like minimizing dependencies cause they've been compromised too often)

alexkeating: I am fine either way, I think our other fields will be pretty small for the entire course of the project. So, manual management shouldn't be too burdensome, but sort-package-json could be good for the long term.

mds1: Will leave this for you guys to decide/resolve, no objections to merging from me!

apbendi (approved):
LGTM and works nice locally too 👍

Issue #16, opened 2023-01-04
UI: "What happens to shielded funds"


wildmolasses:
From reqs: 

> “How to shield” and “What happens to shielded funds” should be obvious to first-time visitors.

Let's include a "What happens to shielded funds" section with appropriate copy. My first thought is that this belongs as a top level component that's visible just beside or below the form, but open to other ideas. Design-wise, should be considered together with #15

PR #31, opened 2023-01-09
Feature/how to section

### LINKED ISSUE #15: UI: "How to shield"
From reqs: 

> “How to shield” and “What happens to shielded funds” should be obvious to first-time visitors.

Let's include a "how to shield" section with appropriate copy. My first thought is that this belongs as a top level component that's visible just beside or below the form, but open to other ideas. Design-wise, should be considered together with #16

alexkeating:
#### Description

- Add how to section under the shielding form

Closes #15 #16 

Screenshot from 2023-01-09 09-25-22

### DIFF
--- a/components/HowTo.tsx
+++ b/components/HowTo.tsx
@@ -7,9 +7,10 @@ const HowTo = () => {
       direction="column"
       border="1px solid"
       borderRadius="0.3rem"
-      padding="1rem"
+      padding="2rem"
       maxWidth="40rem"
       marginTop="3rem"
+      marginBottom="3rem"
     >
       <Heading as="h1" size="lg" alignSelf="center">
         Tutorial
--- a/components/HowTo.tsx
+++ b/components/HowTo.tsx
@@ -1,4 +1,11 @@
-import { Flex, Heading, OrderedList, ListItem, Text } from "@chakra-ui/layout";
+import {
+  Flex,
+  Heading,
+  OrderedList,
+  ListItem,
+  Text,
+  Link,
+} from "@chakra-ui/layout";
 
 // TODO: Copy is a placeholder and pulled from their docs
 const HowTo = () => {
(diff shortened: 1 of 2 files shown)


components/HowTo.tsx
```diff
@@ -0,0 +1,40 @@
...
+      marginBottom="3rem"
+    >
+      <Heading as="h1" size="lg" alignSelf="center">
+        Tutorial
+      </Heading>
+      <Heading as="h4" size="sm" marginTop="1rem">
+        How to shield?
+      </Heading>
+      <OrderedList marginTop="1rem">
+        <ListItem>Create a Railgun wallet</ListItem>
```
mds1, line 22: How do you do this? Can we link to a guide / app that helps with this?

components/HowTo.tsx
```diff
@@ -0,0 +1,40 @@
...
+      padding="2rem"
+      maxWidth="40rem"
+      marginTop="3rem"
+      marginBottom="3rem"
+    >
+      <Heading as="h1" size="lg" alignSelf="center">
+        Tutorial
+      </Heading>
+      <Heading as="h4" size="sm" marginTop="1rem">
+        How to shield?
```
mds1, line 19:
```suggestion
        How to shield funds?
```

alexkeating: I think your understanding is correct here

components/HowTo.tsx
```diff
@@ -0,0 +1,52 @@
...
+      </Heading>
+      <Heading as="h4" size="sm" marginTop="1rem">
+        How to shield funds?
+      </Heading>
+      <OrderedList marginTop="1rem">
+        <ListItem>
+          Create a Railgun wallet on{" "}
+          <Link href="https://app.railway.xyz/" isExternal>
+            Railway
+          </Link>
```
mds1, line 33: One nit is the links currently have no style until you hover, so you can't tell it's a link. Up to you if you want to fix in this PR or track in an issue

image

alexkeating: Yes, good call, I will create a ticket and we can handle styling towards the end. Does that sound good @wildmolasses

alexkeating: #33

mds1 (approved):

PR #32, opened 2023-01-10
Feature/token lists

### LINKED ISSUE #10: Form should react to network change
Spec asks us to support Ethereum, BSC, Polygon, Arbitrum, Goerli. The contents of the token select box should change when network is changed, but that may be a concern of #12. Otherwise, we should probably take care to reset token selection on network change, and in general, no form state from the prior selected network should be persisted.

### LINKED ISSUE #11: Add token lists for Ethereum, BSC, Polygon, Arbitrum, Goerli
Per spec:

> Show searchable list of default ERC20 tokens.
> Support for Ethereum, BSC, Polygon, Arbitrum, Goerli.

Integrate into app!

Will we use coingecko token lists for each network?

alexkeating:
#### Description

- Add global context state for token lists and user balances. This may also close #10  but I cam not sure

Closes #11


alexkeating:
Deploy seems like failing because of the lock file. Working to resolve now.

components/TxForm.tsx
```diff
@@ -27,7 +28,7 @@ const tokenList = [
 
 export const TxForm = () => {
   // TODO: Placeholder notification for shielding
-  //
+  const { tokenList } = useToken();
```
mds1, line 31: We can delete the `const tokenList` example data above

hooks/useTokenBalance.tsx
```diff
@@ -0,0 +1,47 @@
...
+];
+
+const useTokenList = ({ tokenList }: { tokenList: TokenListItem[] }) => {
+  const { chain } = useNetwork();
+  const { address } = useAccount();
+
+  const chainId = chain?.id || 1; // default to mainnet if no chain id
+  const { isLoading, error, data } = useSWR(
+    `userTokenList-${chainId}-${tokenList.length}`,
+    async () => {
```
mds1, line 18: Is my understand of `swr` correct that by default this re-fetches the data every 3 seconds? That seems reasonable to me

hooks/useTokenBalance.tsx
```diff
@@ -0,0 +1,47 @@
...
+      let multicallArgs = [];
+      for (const token of tokenList) {
+        multicallArgs.push({
+          abi: ERC20_ABI,
+          functionName: "balanceOf",
+          address: token.address,
+          args: [address],
+        });
+      }
+      const data = await multicall({
```
mds1, line 31: Do you know what the difference is between wagmi's `multicall` and `readContracts`? Seems the latter uses the former under the hood, but with some other stuff around it. Wondering which we should use. Another option would be [`watchMulticall`](https://wagmi.sh/core/actions/watchMulticall) / [`watchReadContracts`](https://wagmi.sh/core/actions/watchReadContracts), not sure how these various approaches compare

hooks/useTokenBalance.tsx
```diff
@@ -0,0 +1,47 @@
...
+      }
+      let multicallArgs = [];
+      for (const token of tokenList) {
+        multicallArgs.push({
+          abi: ERC20_ABI,
+          functionName: "balanceOf",
+          address: token.address,
+          args: [address],
+        });
+      }
```
mds1, line 30: Nit, but a bit cleaner IMO
```suggestion
      const multicallArgs = tokenList.map(token => {
        abi: ERC20_ABI,
        functionName: "balanceOf",
        address: token.address,
        args: [token.address],
      });
```

hooks/useTokenBalance.tsx
```diff
@@ -0,0 +1,47 @@
...
+      const data = await multicall({
+        contracts: multicallArgs,
+      });
+      const tokelistWithUserBalance = [];
+      for (const [i, token] of tokenList.entries()) {
```
mds1, line 35: Same comment about `map` refactor

public/pancakeswap-top-100.json
```diff
@@ -0,0 +1,797 @@
+{
```
mds1, line 1: Do we need to commit this file? I don't think so since we fetch it in `useTokenList`

utils/networks.ts
```diff
@@ -2,23 +2,29 @@ import { isHexString } from "@ethersproject/bytes";
 
 export type NetworkConfig = {
   blockExplorerUrl: string;
+  tokenList: string;
```
mds1, line 5:
```suggestion
  tokenListUrl: string;
```

package.json
```diff
@@ -20,6 +20,7 @@
     "next": "^13.0.3",
     "react": "^18.1.0",
     "react-dom": "^18.1.0",
+    "swr": "^2.0.0",
```
mds1, line 23: _(arbitrary comment location to start a thread)_

I only see 3 tokens in the dropdown when connected to mainnet, is that expected with this PR?

mds1 (commented):
Looks good!

public/pancakeswap-top-100.json
alexkeating: I tried fetching from github. I was getting some cors errors and thought moving it to the repo would be easier with the idea of potentially moving it to IPFS if we want to remove it from our bundle. It still uses the useTokenList method and fetches the local file.

contexts/TokenContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+export type TokenContextType = {
+  tokenList: TokenListContextItem[];
+  isLoading: boolean;
+};
+const initialContext = {
+  isLoading: false,
+  tokenList: [],
+};
+
+const TokenContext = createContext<TokenContextType>(initialContext);
```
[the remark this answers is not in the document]
alexkeating: TokenContext  has the data from useTokenList with the balance value added to it. It stores the value in a higher level state that can be accessed in lower level components.

We need it for the balance info and the regular hooks don't persist data across components. So, if we need to use token data in multiple parts of the app we would be fetching multiple times. That being said we could combine some of these and even remove the context, but then some of the logic would have multiple responsibilities

package.json
alexkeating: Hmm, that actually seems like the changes were missing as those are the placeholder tokens. Mainnet should have a big list and Goerli will only have 2

public/pancakeswap-top-100.json
mds1: Will a browser always load this file in the initial bundle / would it slow page load times with this approach? Regardless, we can always track removing/optimizing this in a separate issue

contexts/TokenContext.tsx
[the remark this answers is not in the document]
mds1: > We need it for the balance info and the regular hooks don't persist data across components. So, if we need to use token data in multiple parts of the app we would be fetching multiple times.

Wdym by this exactly? My understanding was that `useSWR` handles the http requests to persist the response / avoid duplicate fetches when the hook is used in multiple components, by fetch every 3 seconds and caching results instead (my understanding could be wrong)

hooks/useTokenBalance.tsx
alexkeating: I think the default is 2 seconds, but we can add a custom configuration if that is too often. 

Default configuration: https://github.com/vercel/swr/blob/main/_internal/utils/config.ts#L48

hooks/useTokenBalance.tsx
alexkeating: Under the hood readContracts uses multicall and I think the main difference is that reacContracts adds some error handling functionality. Based on [this](https://github.com/wagmi-dev/wagmi/blob/main/packages/core/src/actions/contracts/readContracts.ts#L21) comment. So, I think I should probably change this to readContracts.

The watch methods look cool. The only benefit would be if we listen to block changes because we currently handle refetching on chain  change.

public/pancakeswap-top-100.json
alexkeating: Hm, I am not sure, but I will make a ticket for this optimization

alexkeating: #38

contexts/TokenContext.tsx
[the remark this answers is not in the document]
alexkeating: I think your understanding is correct, but I think SWR has some internal cache in memory that maintains the data.

What I mean by the statement above is instead of sharing data through passing props down through the component tree we could use context which allows us to skip components in the component tree and share state. Regular hooks will only scope data to the component they are being used in unless they are using context or some other state management solution.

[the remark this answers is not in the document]
alexkeating: That being said, it may be simpler to move balances outside of this context because it is seeming more and more likely it will potentially change frequently

package.json
alexkeating: Is this still an issue?

hooks/useTokenBalance.tsx
mds1: That seems fine for now, can always change it later

hooks/useTokenBalance.tsx
mds1: Cool, yea I don't think we need the watch methods currently. Though `multicall` [also](https://github.com/wagmi-dev/wagmi/blob/2beabc6a587880345f62c2fd305b4c1f1b7339e1/packages/core/src/actions/contracts/multicall.ts#L23-L24) supports the `allowFailure` option, so I'm not sure. 

I just asked about the difference [here](https://github.com/wagmi-dev/wagmi/discussions/1652), tracking in https://github.com/ScopeLift/railgun-shielder/issues/39 so marking this resolved

hooks/useTokenBalance.tsx
```diff
@@ -0,0 +1,45 @@
...
+          abi: ERC20_ABI,
+          functionName: "balanceOf",
+          address: token.address,
+          args: [address],
+        };
+      });
+      const data = await readContracts({
+        contracts: readContractsArgs,
+      });
+      const tokelistWithUserBalance = tokenList.map((token, i) => {
```
mds1, line 33: Typo, feel free to self-resolve once the change is made
```suggestion
      const tokenListWithUserBalance = tokenList.map((token, i) => {
```

contexts/TokenContext.tsx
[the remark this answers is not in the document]
mds1: Got it, appreciate the explanation! Deferring to you in case we need an issue to track that, or if the change will come naturally from other changes we make. Feel free to resolve this comment yourself.

package.json
mds1: Nope, thanks!

image

image

hooks/useTokenBalance.tsx
```diff
@@ -0,0 +1,45 @@
+import { TokenListItem } from "@/hooks/useTokenList";
+import { BigNumber } from "@ethersproject/bignumber";
+import { readContracts } from "@wagmi/core";
+import useSWR from "swr";
+import { useAccount, useNetwork } from "wagmi";
+
+export const ERC20_ABI = [
+  "function balanceOf(address owner) view returns (uint256 balance)",
```
wildmolasses, line 8: per @mds1 we can use a wagmi import here:

```js
import { erc20ABI } from "@wagmi/core";
```

hooks/useTokenBalance.tsx
```diff
@@ -0,0 +1,45 @@
...
+];
+
+const useTokenList = ({ tokenList }: { tokenList: TokenListItem[] }) => {
```
wildmolasses, line 11:
```suggestion
const useTokenBalances = ({ tokenList }: { tokenList: TokenListItem[] }) => {
```

hooks/useTokenBalance.tsx
```diff
@@ -0,0 +1,45 @@
...
+          balance: data[i] as BigNumber | null,
+        };
+      });
+      return tokelistWithUserBalance;
+    }
+  );
+  return { isLoading, error, data };
+};
+
+export default useTokenList;
```
wildmolasses, line 45:
```suggestion
export default useTokenBalances;
```

hooks/useTokenList.tsx
```diff
@@ -0,0 +1,34 @@
+import { networks } from "@/utils/networks";
+import useSWR from "swr";
```
wildmolasses, line 2: thought we were going to prefer react-query, are we going with swr instead?

utils/networks.ts
```diff
@@ -2,23 +2,29 @@ import { isHexString } from "@ethersproject/bytes";
 
 export type NetworkConfig = {
   blockExplorerUrl: string;
+  tokenListUrl: string;
 };
 
 export const networks = {
   1: {
     blockExplorerUrl: "https://etherscan.io/",
+    tokenListUrl: "https://gateway.ipfs.io/ipns/tokens.uniswap.org",
```
wildmolasses, line 11: would any part of us rather store these statically?

hooks/useTokenBalance.tsx
```diff
@@ -0,0 +1,45 @@
...
+      }
+      const readContractsArgs = tokenList.map((token) => {
+        return {
+          abi: ERC20_ABI,
+          functionName: "balanceOf",
+          address: token.address,
+          args: [address],
+        };
+      });
+      const data = await readContracts({
```
wildmolasses, line 30: is it the case that this will make n rpc calls for n tokens? perhaps we can consider [`watchMulticall`](https://wagmi.sh/core/actions/watchMulticall) instead, any thoughts?

wildmolasses (commented):
looking great, a couple questions / comments

hooks/useTokenBalance.tsx
mds1: This is good as-is, check out the conversation here + the linked wagmi discussion where I asked them https://github.com/ScopeLift/railgun-shielder/pull/32#discussion_r1066225006

hooks/useTokenList.tsx
mds1: @alexkeating has experience with react-query, so think we said we're going to try swr on this project since it's supposed to be a bit simpler (and consequently less flexible as app grows) + it's by vercel/next team so should play nicely together

utils/networks.ts
mds1: Good question. AFAIK we don't have a spec on what tokens specifically to support, so I'd say we can leave it as is for now and revisit/simplify in a future PR

hooks/useTokenBalance.tsx
wildmolasses: ah gotcha thanks

hooks/useTokenList.tsx
alexkeating: Oh sorry I didn't mention this, but along with Matt's reasons above I tried setting up react-query with next and it was causing some errors. Since we weren't strongly opposed to SWR I switched over because there was less friction setting it up.

wildmolasses: sg!

utils/networks.ts
alexkeating: #43

hooks/useTokenBalance.tsx
alexkeating: I don't know why but using this abi causes some new type errors. Spent a little time trying to fix, but seems good enough and maybe an improvement we can make later. FYI the type issue was 

```
Type error: Conversion of type '{} & readonly unknown[]' to type 'BigNumber | null' may be a mistake 
because neither type sufficiently overlaps with the other. If this was intentional, convert the expre
ssion to 'unknown' first.

  31 |         return {
  32 |           ...token,
> 33 |           balance: data[i] as BigNumber | null,
     |                    ^
  34 |         };
  35 |       });
  36 |       return tokelistWithUserBalance;
```

alexkeating:
Fixed or responded to outstanding feedback

hooks/useTokenBalance.tsx
```diff
@@ -0,0 +1,45 @@
...
+          balance: data[i] as BigNumber | null,
+        };
+      });
+      return tokenListWithUserBalance;
+    }
+  );
+  return { isLoading, error, data };
+};
+
+export default useTokenBalances;
```
wildmolasses, line 45: nit: can we change the filename to match the export

wildmolasses (approved):
lgtm, 1 small nit

hooks/useTokenBalance.tsx
wildmolasses: it's a nit, let's track in #44 ; all good!

mds1 (approved):
LGTM pending remaining comments from @wildmolasses

Issue #12, opened 2023-01-04
Token dropdown: make it modal, searchable, react to network change


wildmolasses:
From reqs:

> Show searchable list of default ERC20 tokens.

Although we're merely trying to meet the requirement of a searchable token list, that req necessitates a significant component build out for the token select. UI inspo via app.uniswap.org is strongly encouraged. Let's also make sure the token list reacts to network change.

PR #34, opened 2023-01-11
Initial railgun quickstart sdk integration

### LINKED ISSUE #35: Ask for shield private key and store in context

wildmolasses:
Meant to demonstrate the basic flow to compose and execute an erc20 shield, with a bunch of rough edges and shortcuts taken!

Think this resolves #35 , what do you think?


components/TxForm.tsx
```diff
@@ -29,6 +52,108 @@ export const TxForm = () => {
   // TODO: Placeholder notification for shielding
   //
   const { notifyUser } = useNotifications();
+  let abi = [
+    "function transfer(address,uint256) returns (bool)",
+    "function approve(address,uint256) returns (bool)",
+  ];
```
mds1, line 58: Can replace this with `import { erc20ABI } from '@wagmi/core'`: https://wagmi.sh/core/constants/abis

components/TxForm.tsx
```diff
@@ -4,6 +4,29 @@ import { FormControl, FormLabel } from "@chakra-ui/form-control";
...
+} from "@railgun-community/quickstart";
+import {
+  RailgunERC20AmountRecipient,
+  NetworkName,
+  TransactionGasDetailsSerialized,
+  EVMGasType,
+  NETWORK_CONFIG,
+  deserializeTransaction,
+} from "@railgun-community/shared-models";
+import { BigNumber, Signer, ethers, constants } from "ethers";
```
mds1, line 22: I like to avoid importing the full `ethers` object directly and instead import the specific things you need, in this particular case `import type { TransactionRequest } from '@ethersproject/providers'`. Similarly, `import { MaxUint256 } from '@ethersproject/constants'` can be used instead of importing all constants

 The reason is because I don't know how good tree shaking is and don't want to accidentally leave in the full ethers bundle. _In theory_ this shouldn't matter, but just wanted to mention it in case you have any thoughts here. Not saying we must change this, feel free to resolve this comment yourself

components/TxForm.tsx
```diff
@@ -29,6 +52,108 @@ export const TxForm = () => {
...
+      await (signer as Signer).signMessage(
+        getShieldPrivateKeySignatureMessage()
+      )
+    );
+
+    // if (!doErc20Approval) throw "not prepared";
+    // await doErc20Approval();
+
+    // Public wallet to shield from.
+    const fromWalletAddress = address as `0x{string}`;
```
mds1, line 100: Is this type correct? 
image

components/TxForm.tsx
```diff
@@ -29,6 +52,108 @@ export const TxForm = () => {
...
+    // Public wallet to shield from.
+    transactionRequest.from = address;
+    const tx = await signer?.sendTransaction(transactionRequest);
+    console.log(tx);
+    await tx?.wait().then(() => {
+      notifyUser({
+        alertType: "success",
+        message: "Token was shielded successfully",
+      });
+    });
```
mds1, line 154: Let's make sure we're handling all scenarios here: https://docs.ethers.org/v5/single-page/#/v5/api/providers/types/-%23-providers-TransactionResponse

If the tx reverts the promise rejects so we should catch that. We should also get the receipt returned by `tx.wait()` and sanity check the status is 1, not zero

util/walletAddress.ts
```diff
@@ -0,0 +1,18 @@
+import { createRailgunWallet } from "@railgun-community/quickstart";
+import { entropyToMnemonic, hexValue, randomBytes } from "ethers/lib/utils.js";
+
+/**
+ * @notice Tool to generate a random railgun address.
+ * @returns Railgun address (0zk...123).
+ */
+export const randomWalletAddress = async () => {
```
mds1, line 8: What was this for? Seems it's unused anyway, so should we delete it?

mds1 (commented):
I know this is draft still, so only left a few comments

components/TxForm.tsx
wildmolasses: yeah, down for all that; given that transaction handling will likely be relocated and dealt with in another feature i'm inclined to leave as is here with a TODO, what do you think?

util/walletAddress.ts
wildmolasses: sounds good

components/TxForm.tsx
mds1: TODO works for me!

util/walletAddress.ts
wildmolasses: removed, thanks!

components/TxForm.tsx
wildmolasses: added TODO: https://github.com/ScopeLift/railgun-shielder/pull/34/files#diff-809389dcad3e410c420c6a1e2225d2a9a33a40fa577026c2b17ca4e51f73f36dR134

components/TxForm.tsx
wildmolasses: yeah, the quickstart sdk has a specific address type that we need to conform to which is `0x{string}`. Perhaps we should use some kind of [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).

components/TxForm.tsx
wildmolasses: done https://github.com/ScopeLift/railgun-shielder/blob/e933fe01d57c55ab1872747249001094eb43d487/components/TxForm.tsx#L21

components/TxForm.tsx
```diff
@@ -1,34 +1,143 @@
 import useNotifications from "@/hooks/useNotifications";
+import useShieldPrivateKey from "@/hooks/useShieldPrivateKey";
```
mds1, line 2: _(arbitrary comment location to start a thread)_

When I type into the token amount field I get this error

image

components/TxForm.tsx
```diff
@@ -1,34 +1,143 @@
...
 import { Button } from "@chakra-ui/button";
 import { FormControl, FormLabel } from "@chakra-ui/form-control";
 import { Input, InputGroup, InputRightElement } from "@chakra-ui/input";
 import { Box, Flex } from "@chakra-ui/layout";
 import { Select } from "@chakra-ui/select";
+import {
```
mds1, line 8: _(arbitrary comment location to start a thread)_

When I click the WETH token (connect to goerli) it never actually selects as seen in this gif

CleanShot 2023-01-12 at 13 47 17

.vscode/launch.json
```diff
@@ -0,0 +1,28 @@
+{
```
mds1, line 1: _(arbitrary comment location to start a thread)_

> Think this resolves https://github.com/ScopeLift/railgun-shielder/issues/35 , what do you think?

Seems it only asks for my signature the first time in a session, so I'd say yes

.vscode/launch.json
```diff
@@ -0,0 +1,28 @@
...
+  "version": "0.2.0",
```
mds1, line 2: _(arbitrary comment location to start a thread)_

I have WETH in my wallet but am unable to do the shielding. I sign the message, but never get the approve or shield tx prompts. Instead I get this error, using metamask + goerli

image

wildmolasses: the approve is currently commented out

components/TxForm.tsx
```diff
@@ -1,34 +1,143 @@
...
-    logoURI: "https://assets.coingecko…humb/AAVE.png?1601374110",
+    logoURI: "",
   },
 ];
 
 export const TxForm = () => {
-  // TODO: Placeholder notification for shielding
-  //
-  const { notifyUser } = useNotifications();
+  const network = NetworkName.EthereumGoerli;
```
alexkeating, line 45: Maybe create a ticket to add this to the network global state.

.vscode/launch.json
mds1: Can you give more context here? Should that be uncommented out in this PR, or are we handling that in a follow up PR somewhere? (took a quick look and don't see an issue for approvals)

pages/index.tsx
```diff
@@ -1,16 +1,27 @@
...
 const Home: NextPage = () => {
+  const [isProviderLoaded, setProviderLoaded] = useState<Boolean>(false);
+  useEffect(() => {
+    const fn = async () => {
+      await loadProviders();
+      setProviderLoaded(true);
+      console.log(true);
+    };
+    fn();
+  }, []);
```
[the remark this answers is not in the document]
wildmolasses: in this case, we are wanting an assurance that providers have been loaded into railgun. In any case, we should do this more elegantly; tracking in #37

components/TxForm.tsx
wildmolasses: tracking here https://github.com/ScopeLift/railgun-shielder/issues/41

.vscode/launch.json
wildmolasses: i'm uncommenting it, and we will track a more elegant `approve` in #42

components/TxForm.tsx
wildmolasses: should be fixed now

components/TxForm.tsx
wildmolasses: made a small fix so that this is no longer happening
we will be replacing this select component in #12

alexkeating (approved):

.vscode/launch.json
mds1: I know you have #42 for more elegant approve, but so far I haven't successfully shielded anything on this branch which is why I haven't approved yet 😅

mds1: Now that my approval tx went through, I went through the flow again and it does work now! So I guess we can mark this resolved? Will defer to you, but if so consider this approved from me

Issue #17, opened 2023-01-04
Transaction: "Review transaction" modal


wildmolasses:
From reqs:

> Show a “Review transaction” view after selecting the amount(s) and token(s) to shield, which highlights shielding fee.

UI inspo from uniswap

mds1:
> UI inspo from uniswap

Screenshot for reference 

<img width="400" alt="image" src="https://user-images.githubusercontent.com/17163988/210563167-dd0d1a7f-72b7-4b6a-8f40-e04b99771f08.png">

PR #45, opened 2023-01-17
Move approval into a separate button from shield

alexkeating:
#### Description

I did this as a part of #17. Before creating the shield review modal it made sense to have an approval button, and add the review modal that pops up after approval. Instead of doing an approval on every transaction we fetch the allowances on loading of the app and check whether the token needs approval.

### DIFF
--- a/components/TxForm.tsx
+++ b/components/TxForm.tsx
@@ -30,16 +30,6 @@ import {
   useProvider,
 } from "wagmi";
 
-// 1. onClick pop modal
-// 2. show data, and then shield
-//
-// Modal info
-// 1. amount
-// 2. tokens
-// 3. recipient
-// 4. Shielding and gas fees
-//
-// Button should show approval if amount is greater than approval amount
 export const TxForm = () => {
   // TODO: Placeholder notification for shielding
   const { tokenList, tokenAllowances } = useToken();
@@ -66,7 +56,6 @@ export const TxForm = () => {
     ],
   });
   const { writeAsync: doErc20Approval } = useContractWrite(config);
-  // fetch approval amount
   const [recipient, setRecipient] = useState<string>(
     "0zk1qyn0qa5rgk7z2l8wyncpynmydgj7ucrrcczhl8k27q2rw5ldvv2qrrv7j6fe3z53ll5j4fjs9j5cmq7mxsaulah7ykk6jwqna3nwvxudp5w6fwyg8cgwkwwv3g4"
   );
--- a/components/TxForm.tsx
+++ b/components/TxForm.tsx
@@ -199,7 +199,8 @@ export const TxForm = () => {
             if (!doErc20Approval) {
               notifyUser({
                 alertType: "error",
-                message: "Page is not prepared for ERC20 approval",
+                message:
+                  "Page is not prepared for ERC20 approval. Please try again in a few seconds",
               });
               return;
             }
--- a/components/TxForm.tsx
+++ b/components/TxForm.tsx
@@ -20,7 +20,7 @@ import {
   deserializeTransaction,
 } from "@railgun-community/shared-models";
 import { erc20ABI } from "@wagmi/core";
-import { ethers, constants } from "ethers";
+import { ethers, constants, BigNumber } from "ethers";
 import { useState } from "react";
 import {
   useAccount,
@@ -30,9 +30,19 @@ import {
   useProvider,
 } from "wagmi";
 
+// 1. onClick pop modal
+// 2. show data, and then shield
+//
+// Modal info
+// 1. amount
+// 2. tokens
+// 3. recipient
+// 4. Shielding and gas fees
+//
+// Button should show approval if amount is greater than approval amount
 export const TxForm = () => {
   // TODO: Placeholder notification for shielding
-  const { tokenList } = useToken();
+  const { tokenList, tokenAllowances } = useToken();
   const network = NetworkName.EthereumGoerli;
   const { txNotify } = useNotifications();
   let abi = [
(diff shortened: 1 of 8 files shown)


components/TxForm.tsx
```diff
@@ -20,7 +20,7 @@ import {
   deserializeTransaction,
 } from "@railgun-community/shared-models";
 import { erc20ABI } from "@wagmi/core";
-import { ethers, constants } from "ethers";
+import { ethers, constants, BigNumber } from "ethers";
```
mds1, line 23: _(arbitrary comment location to start a thread)_

When I run `yarn` I get a different lockfile, mind updating the lockfile? Think it's just an artifact of the two open PRs that were merge prior to this having different lockfiles

alexkeating: Weird, yes I will do that

hooks/useTokenAllowances.tsx
```diff
@@ -0,0 +1,45 @@
...
+          functionName: "allowance",
+          address: token.address,
+          args: [address, contractAddress],
+        };
+      });
+      const data = await readContracts({
+        contracts: readContractsArgs,
+      });
+      const allowancesPerToken = new Map();
+      const tokenListWithUserBalance = tokenList.map((token, i) => {
```
mds1, line 36: This `tokenListWithUserBalance` value is never used. I think this should be a `.forEach` instead of `.map` if we don't need the return value

But I'm wondering if the `useTokenAllowances` and `useTokenBalances` hooks should just be merged into a `useTokens` hook (or whatever name you like), since we'll always need both, and by merging them we can remove an RPC call

contexts/TokenContext.tsx
```diff
@@ -27,6 +30,12 @@ export const TokenListProvider = ({ children }: { children: ReactNode }) => {
     error: balanceError,
     data,
   } = useTokenBalances({ tokenList: tokenList || [] });
+  const {
+    isLoading: allowanceIsLoading,
+    error: allowanceError,
```
mds1, line 35: `allowanceIsLoading` and `allowanceError` are never used, but seems useful to log if they return an error below like we do for token balances

components/TxForm.tsx
```diff
@@ -190,9 +190,25 @@ export const TxForm = () => {
...
-        Shield
-      </Button>
+      {needsApproval ? (
+        <Button
+          size="lg"
+          mt="1rem"
+          width="100%"
+          onClick={async () => {
+            if (!doErc20Approval) {
+              throw "not prepared";
```
mds1, line 200: Instead of throwing can we wait for the preparation to finish?

components/TxForm.tsx
mds1: I'm still reproducing this, so yea I think that's ok to do

components/TxForm.tsx
alexkeating: Yes, we can should probably pop a notification instead

hooks/useTokenAllowances.tsx
alexkeating: Yes, I think that is a good idea. Do you like the use of a map instead of adding and extra value on a token item in the token list? FYI, I personally prefer the map

components/TxForm.tsx
alexkeating: Deleted and regenerated, hopefully this resolves those issues

hooks/useTokenAllowances.tsx
mds1: > Do you like the use of a map instead of adding an extra value on the token item in the token list? FYI, I personally prefer the map

Sorry not sure I totally follow, do you mind expanding on the two options?

components/TxForm.tsx
alexkeating: Regenerated now

hooks/useTokenAllowances.tsx
alexkeating: Yes, the token list is an array of type TokenListItem, and we build that array by fetching the the token list and then adding the balance values to all the tokens in the token list.

For the allowances we fetch all the allowance values and then create a map with a key of token address and a value of allowance.

My question was if we are going merge both balances and allowances together which data structure would you prefer. And the structures would be as follows

1. An array of tokens from the token list with both the allowance and balance added to each token
2. A map with a key of token address and a value of an object with a key of balance and allowance.

I prefer 2 because it is more efficient to get a single value then 1. Is that clearer?

contexts/TokenContext.tsx
alexkeating: These should be used now

components/TxForm.tsx
mds1: Huh I'm still getting a different lockfile when I run `yarn`. @wildmolasses any thoughts?

hooks/useTokenAllowances.tsx
mds1: Ahh yes I see. Thanks! I'm good with a `Map`, we can still get all keys from the map if needed. @wildmolasses let us know if you have any thoughts/objections here

components/TxForm.tsx
```diff
@@ -190,9 +190,29 @@ export const TxForm = () => {
...
+      {needsApproval ? (
+        <Button
+          size="lg"
+          mt="1rem"
+          width="100%"
+          onClick={async () => {
+            if (!doErc20Approval) {
+              notifyUser({
+                alertType: "error",
+                message: "Page is not prepared for ERC20 approval",
```
mds1, line 202:
```suggestion
                message: "Page is not prepared for ERC20 approval. Please try again in a few seconds",
```

components/TxForm.tsx
alexkeating: Can you post what the yarn.lock changes are? Maybe a pr against this branch.

wildmolasses: are you guys using the same version of yarn? are you both using volta?

components/TxForm.tsx
```diff
@@ -62,6 +62,9 @@ export const TxForm = () => {
   const [tokenAddress, setTokenAddress] = useState<string>();
   const [tokenAmount, setTokenAmount] = useState<string>("");
   const [tokenDecimals, setTokenDecimals] = useState<number>();
+  const needsApproval = ethers.utils
```
wildmolasses, line 65: currently the approval is hardcoded for WETH. shall we address in this PR, or elsewhere?

wildmolasses (commented):
lgtm, one comment on hardcoded token in allowance

components/TxForm.tsx
alexkeating: Yes, I can replace line 50 with tokenAddress

components/TxForm.tsx
mds1: ah yes I was on the wrong yarn version, thanks!

components/TxForm.tsx
alexkeating: Latest commit should handle this

mds1 (approved):

Issue #43, opened 2023-01-16
Potentially store token list statically


alexkeating:
We should revisit how we store the token list once we understand which tokens we need to support. We may want to switch to storing the tokens statically.

wildmolasses:
another consideration here is that we don't support rebase tokens, and want to check token lists against a static blacklist. if we fetch hosted token lists dynamically, there is a risk that a rebase token will have been added, but that our blacklist has not yet been adjusted.

wildmolasses:
See @mds1 comment on ipfs problems

https://github.com/ScopeLift/railgun-shielder/issues/38#issuecomment-1400636356

PR #46, opened 2023-01-18
Shield base token

### LINKED ISSUE #36: Notify on transaction send and on success/fail
Build transaction send flow (blocked by #17) and notify

### LINKED ISSUE #40: Use sdk to shield base token
Currently frontend will only shield ERC20s

See [https://docs.railgun.org/developer-guide/quickstart/transactions/shielding/shield-base-token](docs) on how to shield base token

### LINKED ISSUE #49: Create constants.ts and include native ETH address
> Wdyt of having a `constants.ts` file, which defines all frontend constants—supported chains, ETH address, WETH addresses per network, any ABIs, etc? It's a pattern I've used in frontends and find it very helpful. If we do it, it should be a separate issue / PR

_Originally posted by @mds1 in https://github.com/ScopeLift/railgun-shielder/pull/46#discussion_r1081872410_

wildmolasses:
Closes #40 
Closes #49 
Closes #36

- shields base token
- adds weth, consumable from useToken hook
- refactors the shield tx stuff into a hook

### DIFF
--- a/components/TxForm.tsx
+++ b/components/TxForm.tsx
@@ -10,7 +10,7 @@ import { getRailgunSmartWalletContractForNetwork } from "@railgun-community/quic
 import { NetworkName } from "@railgun-community/shared-models";
 import { erc20ABI } from "@wagmi/core";
 import { ethers, constants, BigNumber } from "ethers";
-import { useState } from "react";
+import React, { useState } from "react";
 import { usePrepareContractWrite, useContractWrite } from "wagmi";
 
 export const TxForm = () => {
@@ -18,10 +18,6 @@ export const TxForm = () => {
   const { tokenList, tokenAllowances } = useToken();
   const network = NetworkName.EthereumGoerli;
   const { txNotify, notifyUser } = useNotifications();
-  let abi = [
-    "function transfer(address,uint256) returns (bool)",
-    "function approve(address,uint256) returns (bool)",
-  ];
   const [tokenAddress, setTokenAddress] = useState<string>();
   const [tokenAmount, setTokenAmount] = useState<string>("");
   const [tokenDecimals, setTokenDecimals] = useState<number>();
@@ -54,7 +50,6 @@ export const TxForm = () => {
       tokenDecimals,
       recipient,
     });
-    await tx?.wait();
     txNotify(tx!.hash);
   };
 
--- a/components/TxForm.tsx
+++ b/components/TxForm.tsx
@@ -10,6 +10,7 @@ import { getRailgunSmartWalletContractForNetwork } from "@railgun-community/quic
 import { NetworkName } from "@railgun-community/shared-models";
 import { erc20ABI } from "@wagmi/core";
 import { ethers, constants, BigNumber } from "ethers";
+import { getAddress } from "ethers/lib/utils.js";
 import React, { useState } from "react";
 import { usePrepareContractWrite, useContractWrite } from "wagmi";
 
@@ -36,9 +37,11 @@ export const TxForm = () => {
   );
   const { shield } = useRailgunTx();
 
-  const needsApproval = ethers.utils
-    .parseUnits(tokenAmount || "0", tokenDecimals)
-    .gt(tokenAllowances.get(tokenAddress || "") || BigNumber.from(0));
+  const needsApproval =
+    tokenAddress !== getAddress(`0x${"e".repeat(40)}`) &&
+    ethers.utils
+      .parseUnits(tokenAmount || "0", tokenDecimals)
+      .gt(tokenAllowances.get(tokenAddress || "") || BigNumber.from(0));
 
   const doSubmit: React.FormEventHandler = async (e) => {
     // TODO: Form validation
--- a/components/TxForm.tsx
+++ b/components/TxForm.tsx
@@ -1,51 +1,27 @@
 import { useToken } from "@/contexts/TokenContext";
 import useNotifications from "@/hooks/useNotifications";
-import useShieldPrivateKey from "@/hooks/useShieldPrivateKey";
+import useRailgunTx from "@/hooks/useRailgunTx";
 import { Button } from "@chakra-ui/button";
 import { FormControl, FormLabel } from "@chakra-ui/form-control";
 import { Input, InputGroup, InputRightElement } from "@chakra-ui/input";
 import { Box, Flex } from "@chakra-ui/layout";
 import { Select } from "@chakra-ui/select";
-import {
-  gasEstimateForShield,
-  gasEstimateForShieldBaseToken,
-  getRailgunSmartWalletContractForNetwork,
-  populateShield,
-  populateShieldBaseToken,
-} from "@railgun-community/quickstart";
-import {
-  RailgunERC20AmountRecipient,
-  RailgunERC20Amount,
-  NetworkName,
-  TransactionGasDetailsSerialized,
-  EVMGasType,
-  NETWORK_CONFIG,
-  deserializeTransaction,
-} from "@railgun-community/shared-models";
+import { getRailgunSmartWalletContractForNetwork } from "@railgun-community/quickstart";
+import { NetworkName } from "@railgun-community/shared-models";
 import { erc20ABI } from "@wagmi/core";
 import { ethers, constants, BigNumber } from "ethers";
 import { useState } from "react";
-import {
-  useAccount,
-  useSigner,
-  usePrepareContractWrite,
-  useContractWrite,
-  useProvider,
-} from "wagmi";
+import { usePrepareContractWrite, useContractWrite } from "wagmi";
 
 export const TxForm = () => {
   // TODO: Placeholder notification for shielding
-  const { tokenList, tokenAllowances, weth } = useToken();
+  const { tokenList, tokenAllowances } = useToken();
   const network = NetworkName.EthereumGoerli;
   const { txNotify, notifyUser } = useNotifications();
   let abi = [
     "function transfer(address,uint256) returns (bool)",
     "function approve(address,uint256) returns (bool)",
   ];
-  const { isConnected, address } = useAccount();
-  const { data: signer } = useSigner();
-  const provider = useProvider();
-  const { getShieldPrivateKey } = useShieldPrivateKey();
   const [tokenAddress, setTokenAddress] = useState<string>();
   const [tokenAmount, setTokenAmount] = useState<string>("");
   const [tokenDecimals, setTokenDecimals] = useState<number>();
@@ -62,6 +38,7 @@ export const TxForm = () => {
   const [recipient, setRecipient] = useState<string>(
     "0zk1qyn0qa5rgk7z2l8wyncpynmydgj7ucrrcczhl8k27q2rw5ldvv2qrrv7j6fe3z53ll5j4fjs9j5cmq7mxsaulah7ykk6jwqna3nwvxudp5w6fwyg8cgwkwwv3g4"
   );
+  const { shield } = useRailgunTx();
 
   const needsApproval = ethers.utils
     .parseUnits(tokenAmount || "0", tokenDecimals)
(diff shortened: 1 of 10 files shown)


alexkeating (commented):
Liking the useRailgunTx hook, noticed one warning that potentially needs resolution

hooks/useTokenList.tsx
```diff
@@ -27,8 +28,18 @@ export const useTokenList = () => {
       const tokenList = json.tokens.filter(
         (token) => token.chainId === chainId
       );
-      return tokenList;
+      const weth = tokenList.find((token) => token.symbol === "WETH")?.address;
+      if (!weth) throw new Error("No WETH found in token list");
```
mds1, line 32: Intead of throwing, we can manually define WETH (per-chain, of course, since it's different per-chain). Need to find what the equivalents are for BSC and Polygon, probably WBNB and WMATIC if I had to guess

hooks/useRailgunTx.tsx
```diff
@@ -0,0 +1,188 @@
...
+  const { weth } = useToken();
+  const provider = useProvider();
+
+  const shield = async (args: {
+    tokenAddress: string;
+    tokenAmount: string;
+    tokenDecimals: number;
+    recipient: string;
+  }) => {
+    if (args.tokenAddress === getAddress(`0x${"e".repeat(40)}`))
```
mds1, line 37: Wdyt of having a `constants.ts` file, which defines all frontend constants—supported chains, ETH address, WETH addresses per network, any ABIs, etc? It's a pattern I've used in frontends and find it very helpful. If we do it, it should be a separate issue / PR

hooks/useRailgunTx.tsx
```diff
@@ -0,0 +1,188 @@
...
+    // Public wallet to shield from.
+    const fromWalletAddress = address as `0x{string}`;
+
+    // The shieldPrivateKey enables the sender to decrypt
+    // the receiver's address in the future.
+    const shieldPrivateKey = await getShieldPrivateKey();
+
+    const wrappedERC20Amount: RailgunERC20Amount = {
+      tokenAddress: weth!, // wETH
+      amountString: parseUnits(tokenAmount!, tokenDecimals).toHexString(), // hexadecimal amount
```
mds1, line 61: I think we can remove the `!` assertion from `tokenAmount`. Also is having the amount as a hex string a requirement of the railgun SDK?

hooks/useRailgunTx.tsx
```diff
@@ -0,0 +1,188 @@
...
+    };
+
+    const { gasEstimateString, error: err } =
```
mds1, line 64: Similarly, does that mean `gasEstimateString` is also a hex string, and not a decimal string? Accepting inputs as ethers.js' `BigNumberish` type and returning `BigNumbers` or `bigint` would be a much nicer SDK interface IMO

hooks/useRailgunTx.tsx
```diff
@@ -0,0 +1,188 @@
...
+    if (err) {
+      throw err;
+    }
+
+    const { maxFeePerGas, maxPriorityFeePerGas } = await provider.getFeeData();
+
+    const gasDetailsSerialized: TransactionGasDetailsSerialized = {
+      evmGasType: EVMGasType.Type2, // Depends on the chain (BNB uses type 0)
+      gasEstimateString: gasEstimateString!, // Output from gasEstimateForShieldBaseToken
+      maxFeePerGasString: maxFeePerGas!.toHexString(), // Current gas Max Fee
```
mds1, line 81: Do we have an issue tracking all the changes we need when we convert from just Goerli support to supporting all chains? Seems the single-chain assumptions are hardcoded and spread across various locations, worried it may get a bit tedious to refactor. Probably fine though

hooks/useRailgunTx.tsx
```diff
@@ -0,0 +1,188 @@
...
+    if (error) {
+      throw error;
+    }
+
+    const { chain } = NETWORK_CONFIG[network];
+
+    const transactionRequest: ethers.providers.TransactionRequest =
+      deserializeTransaction(
+        serializedTransaction!,
+        undefined, // nonce (optional)
```
mds1, line 101: Another SDK comment: Having an optional second arg but a required third arg feels a bit clunky. An object with optional fields seems cleaner

hooks/useRailgunTx.tsx
```diff
@@ -0,0 +1,188 @@
...
+        chain.id
+      );
+
+    // Public wallet to shield from.
+    transactionRequest.from = address;
```
mds1, line 106: Another SDK comment that we discussed—this line should not be necessary, we should just pass this address into `deserializeTransaction` and get the final `transactionRequest` object back. (Leaving these SDK comments here to see if you agree, then we can port them over to our doc)

hooks/useRailgunTx.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+    const transactionRequest: ethers.providers.TransactionRequest =
+      deserializeTransaction(
+        serializedTransaction as string,
+        undefined, // nonce (optional)
+        chain.id
+      );
+
+    // Public wallet to shield from.
+    transactionRequest.from = address;
```
mds1, line 181: Feels like a lot of duplicate code between the native token and ERC20 methods. Wonder if we can/should dedupe a bit?

contexts/TokenContext.tsx
```diff
@@ -10,17 +10,19 @@ type TokenListContextItem = TokenListItem & { balance: BigNumber | null };
 export type TokenContextType = {
   tokenList: TokenListContextItem[];
   isLoading: boolean;
+  weth?: string;
```
mds1, line 13: Personally not a fan of this typing as it's not obvious this `string` should be an address. I'd suggest making the type `TokenInfo` (from the token list, you can get that type from the uniswap token lists package), or renaming these to things like `wethAddress`

components/TxForm.tsx
```diff
@@ -39,13 +22,6 @@ export const TxForm = () => {
...
-  console.log(
-    getRailgunSmartWalletContractForNetwork(NetworkName.EthereumGoerli).address
-  );
-  const { isConnected, address } = useAccount();
-  const { data: signer } = useSigner();
-  const provider = useProvider();
-  const { getShieldPrivateKey } = useShieldPrivateKey();
   const { config } = usePrepareContractWrite({
     address: "0xb4fbf271143f4fbf7b91a5ded31805e42b2208d6",
     abi: erc20ABI,
```
mds1, line 27: Above, the `abi` variable is declared but never used

components/TxForm.tsx
```diff
@@ -62,77 +38,20 @@ export const TxForm = () => {
...
-        chain.id
-      );
-
-    // Public wallet to shield from.
-    transactionRequest.from = address;
-    // TODO: handle transaction status and notifications in a feature
-    const tx = await signer?.sendTransaction(transactionRequest);
-    await tx?.wait().then(() => {
-      txNotify(tx.hash);
+    if (!tokenAddress || !tokenAmount || !tokenDecimals || !recipient)
```
mds1, line 45: Is checking both `tokenDecimals` and `tokenAddress` redundant? Since they're both populated from the token dropdown

mds1 (commented):
UI is working well!

hooks/useRailgunTx.tsx
wildmolasses: i'm down!

wildmolasses: tracking in #49

hooks/useRailgunTx.tsx
wildmolasses: > Also is having the amount as a hex string a requirement of the railgun SDK?

yes!

hooks/useRailgunTx.tsx
wildmolasses: good call! currently inputs must be hex strings.

hooks/useRailgunTx.tsx
wildmolasses: i wrote an issue #48 for the gas type

hooks/useRailgunTx.tsx
wildmolasses: I looked into deduping, it got a bit convoluted. For simplicity i thought we might PR this version, and if we see the opportunity to refactor then we could go for it. What do you think?

components/TxForm.tsx
```diff
@@ -62,77 +38,20 @@ export const TxForm = () => {
...
+      throw new Error("bad form");
+    const tx = await shield({
+      tokenAddress,
+      tokenAmount,
+      tokenDecimals,
+      recipient,
     });
+    await tx?.wait();
+    txNotify(tx!.hash);
```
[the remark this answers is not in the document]
wildmolasses: handled 14e08c7

hooks/useTokenList.tsx
wildmolasses: we could do that, but don't we expect WETH in each token list? and if it's not there, mightn't it be grounds for a big error?

hooks/useRailgunTx.tsx
wildmolasses: `!` removed c6cccf5

hooks/useRailgunTx.tsx
wildmolasses: agreed!

hooks/useRailgunTx.tsx
wildmolasses: yeah feels a lil neater

hooks/useTokenList.tsx
mds1: Hmm, yea in theory each token list should have the wrapped version of the native token (WETH), and we can verify that upfront before committing to that token list. Partly related to https://github.com/ScopeLift/railgun-shielder/issues/43, because if the list is a static, local JSON file this line can neve occur. But if loading the token lists from a URL, it's possible the list changes and removes WETH for whatever reason. In that case we _can_ throw an error, but since we know the address of an immutable WETH contract that will always work, we can just plug that in instead

contexts/TokenContext.tsx
wildmolasses: made it a token item 37508e2

components/TxForm.tsx
wildmolasses: fixed!

hooks/useRailgunTx.tsx
wildmolasses: update: i just made the constants file in 1836e62

hooks/useRailgunTx.tsx
mds1: Sounds great, thanks!

hooks/useTokenList.tsx
```diff
@@ -27,8 +28,18 @@ export const useTokenList = () => {
...
+      const eth: TokenListItem = {
+        chainId,
+        symbol: "ETH",
+        address: ethers.utils.getAddress(`0x${"e".repeat(40)}`),
+        decimals: 18,
+        name: "ETH",
+        logoURI: "",
+      };
+      return { tokenList: [eth, ...tokenList], weth };
```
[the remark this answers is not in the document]
wildmolasses: this commit gets balances for eth and includes them in the whole getTokenBalances thing f04d61e

components/TxForm.tsx
wildmolasses: perhaps redundant. the alternative is assume the token dropdown is fulfilling its duties.

and typescript won't let me!

Screenshot 2023-01-19 at 11 42 07 PM

hooks/useTokenList.tsx
wildmolasses: ok, how's this for ya! e7051b8

alexkeating (approved):

hooks/useTokenList.tsx
mds1: Like it! Just needs a few fixes.

First, all the WETH addresses are `0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2` but that's not correct. For example, `cast code 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 --rpc-url $POLYGON_RPC_URL`(or another RPC) shows there is no code at those addresses. Usually you know you've found the right contract because it (1) holds a ton of the native token, and (2) is written in solidity 0.4.18 (or similar) since that's what the original mainnet WETH contract is written in, and everyone copies it

Second, not all of them are called WETH, since WETH = Wrapped Ether and other chains have different native token names, e.g. on matic it's Wrapped Matic / WMATIC

Here's the info you need for each chain, in the format "Name / Symbol / Decimals / Address"
- Mainnet: Wrapped Ether / WETH / 18 / 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2
- Goerli: Wrapped Ether / WETH / 18 / 0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6 
- BSC: Wrapped BNB / WBNB / 18 / 0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c
- Polygon: Wrapped Matic / WMATIC / 18 / 0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270
- Arbitrum: Wrapped Ether / WETH / 18 / 0x82aF49447D8a07e3bd95BD0d56f35241523fBab1

components/TxForm.tsx
mds1: Ah interesting, I guess you could change that to `tokenDecimals!` instead. Will defer to whichever you prefer, feel free to resolve this comment yourself.

hooks/useTokenList.tsx
wildmolasses: thanks matt will be fixing in #51

wildmolasses: going to handle in #51 , thank you!

PR #51, opened 2023-01-20
load fallback rpcs for most networks; add wagmi provider as railgun provider; responsive gas type per network

### LINKED ISSUE #37: load provider from wagmi into railgun engine

### LINKED ISSUE #41: Artfully link wagmi network with railgun network to ensure parity
There are places where we need to supply a network value to a railgun sdk method. In TxForm we are hardcoding it to a goerli value. Let's instead supply a value that's conditional on the wagmi network chain value. Perhaps we could write a hook that makes this easy, or access via some context.

### LINKED ISSUE #48: Make gas type conditional on network
aka binance chain should use gas type 0

wildmolasses:
closes #48 
closes #41 
closes #37


utils/networks.ts
```diff
@@ -9,6 +11,9 @@ export type NetworkConfig = {
   railgunNetworkName: NetworkName;
   chainId: number;
   wethAddress: string;
+  evmGasType: EVMGasType;
+  baseToken: { symbol: string; name: string };
+  fallbackProviders: FallbackProviderJsonConfig;
```
alexkeating, line 16: (arbitrary place for a thread)

Noticing this error on polygon when I try to shield. And also should the native token be MATIC rather than Polygon. I wonder if that is an SDK issue

Screenshot from 2023-01-20 11-01-15

hooks/useRailgunProvider.tsx
```diff
@@ -0,0 +1,28 @@
+// import { loadProviders } from "@/utils/railgun";
```
alexkeating, line 1: This should probably be removed

alexkeating (commented):
Only tested on Polygon and Goerli. Goerli looks good, but Polygon didn't seem to work with the native token

utils/networks.ts
```diff
@@ -18,34 +23,124 @@ export const networks = {
...
+    evmGasType: EVMGasType.Type2,
+    baseToken: {
+      symbol: "ETH",
+      name: "Ether",
+    },
+    fallbackProviders: {
+      chainId: mainnet.id,
+      providers: [
+        {
+          provider: "https://cloudflare-eth.com/",
```
mds1, line 35: Should we parameterize all these RPC URLs as env vars or something? For a production app we probably don't want to rely on public RPCs since they're often flaky/rate-limited

utils/networks.ts
```diff
@@ -18,34 +23,124 @@ export const networks = {
...
   [goerli.id]: {
     blockExplorerUrl: "https://goerli.etherscan.io/",
     tokenListUrl: "https://gateway.ipfs.io/ipns/tokens.uniswap.org",
     railgunNetworkName: NetworkName.EthereumGoerli,
     chainId: goerli.id,
     wethAddress: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
+    evmGasType: EVMGasType.Type2,
+    baseToken: {
+      symbol: "GoerliETH",
+      name: "Goerli Ether",
```
mds1, line 56: Should be the same as mainnet: symbol of `ETH` and name of `Ether`

utils/networks.ts
```diff
@@ -18,34 +23,124 @@ export const networks = {
...
+        {
+          provider: "https://rpc.ankr.com/eth_goerli",
+          priority: 1,
+          weight: 1,
+        },
+      ],
+    },
   },
   [bsc.id]: {
     blockExplorerUrl: "https://bscxplorer.com/",
```
mds1, line 75: Recommend using Etherscan's https://bscscan.com/ instead

hooks/useTokenList.tsx
```diff
@@ -34,18 +33,18 @@ export const useTokenList = () => {
         symbol: "WETH",
         address: network.wethAddress,
         decimals: 18,
-        name: "wETH",
+        name: "Wrapped ETH",
```
mds1, line 36: Migrating this comment as a reminder to verify it's been resolved: https://github.com/ScopeLift/railgun-shielder/pull/46#discussion_r1082733840

hooks/useTokenList.tsx
```diff
@@ -34,18 +33,18 @@ export const useTokenList = () => {
...
-      const eth: TokenListItem = {
+      const baseToken: TokenListItem = {
         chainId,
-        symbol: "ETH",
+        symbol: network.baseToken.symbol,
         address: ethAddress,
         decimals: 18,
-        name: "ETH",
+        name: network.baseToken.name,
         logoURI: "",
```
mds1, line 45: Change empty string to `network.baseToken.logoURI`?

hooks/useRailgunProvider.tsx
wildmolasses: thanks! 26011e3

utils/networks.ts
wildmolasses: thanks, we've got the base token names in our network object, changed to MATIC: 2220e8e

will dig into the polygon error

mds1 (commented):
Haven't tested the app yet, just looked at the code so far

hooks/useRailgunTx.tsx
```diff
@@ -73,14 +98,9 @@ const useRailgunTx = () => {
...
-    const { maxFeePerGas, maxPriorityFeePerGas } = await provider.getFeeData();
-
-    const gasDetailsSerialized: TransactionGasDetailsSerialized = {
-      evmGasType: EVMGasType.Type2, // Depends on the chain (BNB uses type 0)
-      gasEstimateString: gasEstimateString!, // Output from gasEstimateForShieldBaseToken
-      maxFeePerGasString: maxFeePerGas!.toHexString(), // Current gas Max Fee
-      maxPriorityFeePerGasString: maxPriorityFeePerGas!.toHexString(), // Current gas Max Priority Fee
-    };
+    const gasDetailsSerialized = await getGasDetailsSerialized(
+      gasEstimateString!
```
mds1, line 102: How do we know `gasEstimateString` will never be undefined?

hooks/useRailgunTx.tsx
```diff
@@ -149,13 +169,9 @@ const useRailgunTx = () => {
...
 
-    const { maxFeePerGas, maxPriorityFeePerGas } = await provider.getFeeData();
-    const gasDetailsSerialized: TransactionGasDetailsSerialized = {
-      evmGasType: EVMGasType.Type2, // Depends on the chain (BNB uses type 0)
-      gasEstimateString: gasEstimateString!, // Output from gasEstimateForShield
-      maxFeePerGasString: maxFeePerGas!.toHexString(), // Current gas Max Fee
-      maxPriorityFeePerGasString: maxPriorityFeePerGas!.toHexString(), // Current gas Max Priority Fee
-    };
+    const gasDetailsSerialized = await getGasDetailsSerialized(
+      gasEstimateString!
```
mds1, line 173: Same question

hooks/useRailgunTx.tsx
wildmolasses: a few lines above (~L98), we check if there was an error in the gas estimation and throw if there was. was thinking this is sufficient.

hooks/useTokenList.tsx
wildmolasses: should be resolved, can you check it out?

mds1: The addresses in `network.ts` look right! I'll mark this resolved

hooks/useRailgunTx.tsx
mds1: I'm not familiar enough with the railgun SDK to know for sure. Is it true that `gasEstimateString` can never be undefined when there is no error? If so this lgtm

utils/networks.ts
wildmolasses: tracking in #55

utils/networks.ts
wildmolasses: Goodbye, Goerli ETH 7cd7d57

utils/networks.ts
wildmolasses: a9209f8

hooks/useTokenList.tsx
wildmolasses: 561c2cd

hooks/useRailgunTx.tsx
wildmolasses: i've made a note in the feedback doc about this, but i feel like for now it's safe to assume

hooks/useRailgunTx.tsx
```diff
@@ -39,6 +42,28 @@ const useRailgunTx = () => {
     return shieldToken(args);
   };
 
+  const getGasDetailsSerialized = async (
+    gasEstimateString: string
```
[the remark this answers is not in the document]
wildmolasses: tracking in #56

utils/networks.ts
mds1: We need to enforce a 30gwei min priority fee in polygon, which is required by the network. Looks like we're setting a 1.5gwei max priority fee 
image

mds1: - default `PriceLimit` is 30 gwei: https://github.com/maticnetwork/bor/blob/d6899d79ea4c4d169d9dde613a9f4b0f1dccc3d8/builder/files/config.toml#L44
- the `validateTx` method returns tx underpriced error when `tx.gasTipCapU256` is less than that value https://github.com/maticnetwork/bor/blob/d6899d79ea4c4d169d9dde613a9f4b0f1dccc3d8/core/tx_pool.go#L769-L772

wildmolasses: e1ce4da 30 gwei min max enforced

hooks/useRailgunProvider.tsx
mds1: `loadProviders` is now used, I'd say we can resolve this

hooks/useRailgunProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const network = networks[chain?.id || 1];
+  const provider = useProvider();
+
+  useEffect(() => {
+    const fn = async () => {
+      setProviderLoaded(false);
+      await loadProviders();
+      setProviderLoaded(true);
+    };
+    fn();
```
mds1, line 19: Do we need a `.catch` or any error handling here, in case `loadProviders` errors?

hooks/useRailgunTx.tsx
```diff
@@ -10,23 +10,29 @@ import {
...
-  NetworkName,
   TransactionGasDetailsSerialized,
   EVMGasType,
   NETWORK_CONFIG,
+  NetworkName,
   deserializeTransaction,
 } from "@railgun-community/shared-models";
 import { ethers } from "ethers";
-import { getAddress, parseUnits } from "ethers/lib/utils.js";
+import { parseUnits } from "ethers/lib/utils.js";
```
mds1, line 20: For some reason VSCode always autocompletes the import to this JS file, but typically it's this
```suggestion
import { parseUnits } from "@ethersproject/units";
```

hooks/useRailgunTx.tsx
```diff
@@ -39,6 +45,41 @@ const useRailgunTx = () => {
...
+        // polygon maxPriorityFee must be at least 30
+        if (network === NetworkName.Polygon) {
+          const gwei30 = parseUnits("30", "gwei");
+          return {
+            ...data,
+            maxPriorityFeePerGas: data.maxPriorityFeePerGas!.lt(gwei30)
+              ? gwei30
+              : data.maxPriorityFeePerGas,
+          };
+        }
```
mds1, line 63: Let's delete this (and any other related) special case polygon handling: we don't currently support polygon as discussed, and when we do support, this will be wrong anyway

hooks/useRailgunProvider.tsx
wildmolasses: if `loadProviders` errors, i'd rather the error bubble up and throw at the top level

hooks/useRailgunTx.tsx
mds1: Yea I'm fine with

```typescript
if (network === NetworkName.Polygon) {
 // TODO
}
```

or something like that. But right now the code makes it seem like the current implementation is correct, which is why I think we should remove it

wildmolasses: will do, thanks!

wildmolasses: 2d9ced1

hooks/useRailgunTx.tsx
```diff
@@ -149,13 +187,11 @@ const useRailgunTx = () => {
...
-      evmGasType: EVMGasType.Type2, // Depends on the chain (BNB uses type 0)
-      gasEstimateString: gasEstimateString!, // Output from gasEstimateForShield
-      maxFeePerGasString: maxFeePerGas!.toHexString(), // Current gas Max Fee
-      maxPriorityFeePerGasString: maxPriorityFeePerGas!.toHexString(), // Current gas Max Priority Fee
-    };
+    const gasDetailsSerialized = await getGasDetailsSerialized(
+      gasEstimateString!
+    );
+
+    console.log(gasDetailsSerialized);
```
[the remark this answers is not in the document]
wildmolasses: ae032f4

hooks/useRailgunTx.tsx
wildmolasses: 0c23838

mds1 (approved):

PR #57, opened 2023-01-20
Fix Approval Bug

alexkeating:
#### Description

- In the token allowance hook we weren't excluding the native token when we were building the map. So, allowances were not being mapped to the correct token.


wildmolasses (approved):
lgtm, thanks @alexkeating

Issue #50, opened 2023-01-20
Store serialized fees in context


wildmolasses:

wildmolasses:
closing in favor of #61

PR #59, opened 2023-01-23
Add review transaction modal

### LINKED ISSUE #17: Transaction: "Review transaction" modal

alexkeating:
#### Description

- Add review modal for shielding transaction

Closes #17 

[Screencast from 2023-01-22 20-48-22.webm](https://user-images.githubusercontent.com/12705146/214108683-d016815d-5c39-48bc-8220-a7ef3b348033.webm)


components/ReviewTransactionModal.tsx
```diff
@@ -0,0 +1,154 @@
...
+                </Heading>
+                <Text size="sm">
+                  {ethers.utils.formatUnits(feeAmount, tokenDecimals)}{" "}
+                  {tokenSymbol}
+                </Text>
+              </Flex>
+            </Flex>
+            <Flex align="center" justify="space-between" marginTop="1.5rem">
+              <Heading size="sm" paddingX={2}>
+                Amount + Fee
```
wildmolasses, line 131: I believe the way this works is that fee is taken out of amount. In other words, if you send 1 ETH, and the fee % is 5%, then actually, you'll be sending .95 ETH, with the .05 ETH fee being deducted.

hooks/useRailgunTx.tsx
```diff
@@ -28,15 +29,24 @@ const useRailgunTx = () => {
...
 
   const shield = async (args: {
     tokenAddress: string;
     tokenAmount: string;
     tokenDecimals: number;
     recipient: string;
   }) => {
-    if (args.tokenAddress === ethAddress) return shieldBaseToken(args);
-    return shieldToken(args);
+    if (args.tokenAddress === ethAddress) {
```
wildmolasses, line 40: might change this block to:

```ts
      setIsShielding(true);
      const resp = args.tokenAddress === ethAddress ? await shieldBaseToken(args) : await shieldToken(args);
      setIsShielding(false);
      return resp;
```

utils/constants.ts
```diff
@@ -1,3 +1,5 @@
 import { getAddress } from "@ethersproject/address";
+import { ethers, constants, BigNumber } from "ethers";
 
 export const ethAddress = getAddress(`0x${"e".repeat(40)}`);
+export const feePercent = BigNumber.from("15").div("100");
```
wildmolasses, line 5: this fee can change and depends on network, but for this PR I'd say this is sufficient! 👍

wildmolasses (changes requested):
Excellent work! Got a couple comments. Also, the video screencast in the PR description is awesome but fyi does not show the transaction modal functionality.

components/ReviewTransactionModal.tsx
```diff
@@ -0,0 +1,154 @@
+import useNotifications from "@/hooks/useNotifications";
+import useRailgunTx from "@/hooks/useRailgunTx";
+import useShieldPrivateKey from "@/hooks/useShieldPrivateKey";
+import { shortenAddress } from "@/utils/address";
+import { feePercent } from "@/utils/constants";
```
mds1, line 5: Lots of unused imports/vars in this file and some others

components/ReviewTransactionModal.tsx
```diff
@@ -0,0 +1,154 @@
...
+  isOpen,
+  onClose,
+  tokenAddress,
+  recipient,
+  tokenAmount,
+  tokenDecimals,
+  tokenSymbol,
+  tokenName,
+}: ReviewTransactionModalProps) => {
+  const network = NetworkName.EthereumGoerli; // TODO: Remove once we support multiple chains
```
mds1, line 55: We should be able to remove this once we rebase against `main`

components/ReviewTransactionModal.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const doSubmit: React.FormEventHandler = async (e) => {
+    // TODO: Form validation
+    if (!tokenAddress || !tokenAmount || !tokenDecimals || !recipient)
+      throw new Error("bad form");
+    const tx = await shield({
+      tokenAddress: tokenAddress,
+      tokenAmount: tokenAmount,
+      tokenDecimals: tokenDecimals,
+      recipient: recipient,
+    });
```
mds1, line 75: Nit, unless there's any objections, let's use the abbreviated form since key and value names match
```suggestion
    const tx = await shield({ tokenAddress, tokenAmount, tokenDecimals, recipient });
```

utils/address.ts
```diff
@@ -0,0 +1,6 @@
+export const shortenAddress = (address: string) => {
+  if (address.length < 9) {
```
mds1, line 2: Would this ever be true?

utils/address.ts
```diff
@@ -0,0 +1,6 @@
...
+    return address;
+  }
+  return `${address.slice(0, 4)}...${address.slice(-5)}`;
```
mds1, line 5: Because all addresses have a `0z` prefix we probably should show the first 4 and last 4 characters, *excluding* the prefix. So `address.slice(0, 6)}...${address.slice(-4)

mds1 (commented):
didn't get much time to review, will continue later today

alexkeating:
> Excellent work! Got a couple comments. Also, the video screencast in the PR description is awesome but fyi does not show the transaction modal functionality.

Ah sorry, replaced it with the correct video

components/ReviewTransactionModal.tsx
alexkeating: Ah okay, didn't know that, I will remove this Amount + fee line item from the modal

utils/constants.ts
alexkeating: Cool, I will add a TODO and create a ticket so this isn't lost

alexkeating: #61

components/ReviewTransactionModal.tsx
alexkeating: Going to add eslint:recommended to the eslint.json so we catch these in the build

utils/address.ts
alexkeating: It would only be true if a developer used this method incorrectly. Maybe we should throw inside this check instead.

utils/constants.ts
wildmolasses: thanks, i will close #50 in favor of #61, because it's worded better :+1:

utils/address.ts
mds1: Why length of 9 though? That seems arbitrary. Shouldn't the check instead be that the address starts with `0zk` and the rest is valid base58? (or whatever format railgun addresses use)

utils/address.ts
mds1: Actually I think the prefix is `0zk`, so we'll need `(0,7)` as the first slice

components/ReviewTransactionModal.tsx
```diff
@@ -0,0 +1,125 @@
+import useNotifications from "@/hooks/useNotifications";
```
mds1, line 1: _(arbitrary comment location to start a thread)_

I'm seeing this error in the console 
image

mds1 (commented):
lgtm overall, two more small comments

utils/address.ts
alexkeating: Yes, I think the 9 was supposed to be like minimum length of the string without repeating characters before we changed the slice parameters. I agree we should validate the the address. I will add that and throw an error if it is an invalid string

utils/address.ts
alexkeating: This should be fixed now

components/ReviewTransactionModal.tsx
alexkeating: This has been removed

components/ReviewTransactionModal.tsx
mds1: This happens as soon as I open the app. Tracking in https://github.com/ScopeLift/railgun-shielder/issues/63, we'll see if #62 fixes it and if not can revisit in a separate PR

mds1 (approved):

wildmolasses (approved):
lgtm

PR #62, opened 2023-01-24
Make token list static

### LINKED ISSUE #43: Potentially store token list statically

alexkeating:
#### Description

* Merge pancakswap list with uniswap list and store statically

Closes #43


public/tokenlist.json
```diff
@@ -0,0 +1,8611 @@
+{
+  "name": "Uniswap Labs Default",
```
mds1, line 2: open to other names, let's just clarify this is now a custom list though. Similarly we should change the version to 1.0.0, currently it's 7.0.0
```suggestion
  "name": "Railgun Shielder",
```

hooks/useTokenList.tsx
```diff
@@ -23,7 +23,7 @@ export const useTokenList = () => {
   const { isLoading, error, data } = useSWR(
     `userTokenList-${chainId}`,
     async () => {
-      const resp = await fetch(network.tokenListUrl);
+      const resp = await fetch("tokenlist.json");
```
mds1, line 26: This doesn't need to be async anymore, and probably doesn't need useSWR either? Or perhaps SWR is worth it since it's still a big file and a lot to process, to avoid re-processing when the component re-renders?

wildmolasses: hmm, after first request it should be cached by browser.
we shouldn't have to fetch this again when the chainId changes, because all chains read from the same list. maybe we could store the initial load to state, and then reactively filter when the network changes.

wildmolasses (changes requested):
looks great, pending some stuff around tokenList loading/switching

public/tokenlist.json
alexkeating: Cool, and I will add a description field with the versions of the lists we combined

hooks/useTokenList.tsx
alexkeating: Should we use useSWRImmutable instead?

https://swr.vercel.app/blog/swr-v1#immutable-mode

mds1: ah yes forgot about that, yea we should

alexkeating: Cool, I split out the fetch and used the immutable hook.

mds1: sweet, but why do we still do `await fetch("tokenlist.json")` instead of just importing the JSON? seems unnecessary to `fetch` it and have two async calls here

hooks/useTokenList.tsx
```diff
@@ -1,6 +1,6 @@
 import { ethAddress } from "@/utils/constants";
 import { networks } from "@/utils/networks";
-import useSWR from "swr";
+import useSWRImmutable from "swr/immutable";
```
mds1, line 3: _(arbitrary comment location to start a thread)_

Seeing these errors in the console, but I was able to shield funds still. Showed up on both mainnet and goerli, haven't tried other networks

image

hooks/useTokenList.tsx
alexkeating: Fetching prevents it from being in the bundle the difference  isn't huge though. Running build with a fetch vs importing the json file looks like below.

With import 
Screenshot from 2023-01-24 15-28-06

With fetch
Screenshot from 2023-01-24 15-28-34

alexkeating: I am open to importing if everyone is good with  little bundle bloat.

mds1: > Fetching prevents it from being in the bundle the difference isn't huge though

Ah interesting.

Personally I prefer the simpler code and 5kb extra of bundle size, since that's pretty negligible compared to a 2.2MB bundle, and that list need to be loaded to use the app no matter what anyway

hooks/useTokenList.tsx
alexkeating: I think this is coming from the loading the ArbitrumGoerli network. Going to remove and we can add back

alexkeating: #64

wildmolasses (approved):
lgtm

mds1 (approved):

PR #65, opened 2023-01-25
Feature/token selection modal

### LINKED ISSUE #12: Token dropdown: make it modal, searchable, react to network change

alexkeating:
### Description

- Add a token selection modal. This should search by both address and name. I didn't add native token logos and the form does not clear on network change. I am going to create a form ticket for clearing state and validation so we can track them.

Closes #12 

[Screencast from 2023-01-25 00-04-13.webm](https://user-images.githubusercontent.com/12705146/214484236-b5b27791-af6a-4c08-967d-e0c10d42cac7.webm)


components/TokenInput.tsx
```diff
@@ -0,0 +1,44 @@
+import TokenSelectionModal from "@/components/TokenSelectionModal";
+import { useToken } from "@/contexts/TokenContext";
```
alexkeating, line 2: #66 Form state clearing ticket

components/TokenInput.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { TokenListContextItem } from "@/contexts/TokenContext";
+import { Input } from "@chakra-ui/input";
+import { useDisclosure } from "@chakra-ui/react";
+import { useState } from "react";
+
+type TokenInputProps = {
+  onSelect: (token: TokenListContextItem) => void; // eslint-disable-line no-unused-vars
```
wildmolasses, line 9: if we extend from an eslint typescript-specific config, then this eslint-disable-line should go away after a vscode window reload.

.eslint.config.json
```ts
{
  "extends": [
    "next/core-web-vitals",
    "eslint:recommended",
    "plugin:@typescript-eslint/eslint-recommended"
  ]
}

**Review by wildmolasses — COMMENTED** [review]
lgtm, one nit on the 3 `eslint-disable-line`s

**Review by mds1 — COMMENTED** [review]
works well!

File: utils/constants.ts [hunk 3]
```diff
@@ -2,6 +2,7 @@ import { getAddress } from "@ethersproject/address";
 import { BigNumber } from "ethers";
 
 export const ethAddress = getAddress(`0x${"e".repeat(40)}`);
+export const ipfsDomain = "https://gateway.ipfs.io/ipfs/";
```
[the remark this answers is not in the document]
alexkeating: Thanks!

components/TxForm.tsx
```diff
@@ -68,28 +68,15 @@ export const TxForm = () => {
...
       <FormControl>
         <FormLabel>Token</FormLabel>
-        <Select
-          size="lg"
-          height="4rem"
-          mb=".75rem"
-          onChange={(e) => {
-            const { address, decimals, symbol, name } =
-              tokenList[+e.target.value];
+        <TokenInput
```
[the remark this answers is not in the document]
alexkeating: I think these fit in and have added them

components/TokenSelectionModal.tsx
```diff
@@ -82,8 +82,9 @@ const TokenSelectionModal = (props: TokenSelectionModalProps) => {
   const { tokenList } = useToken();
   const [searchTerm, setSearchTerm] = useState("");
   const options = {
-    includeScore: false,
-    keys: ["address", "name"],
+    includeScore: true,
+    keys: ["address", "name", "symbol"],
+    threshold: 0.2,
```
mds1, line 87: Just curious how you chose this threshold? (0.2 is totally fine with me, just wondering if it's based on testing, past usage of fuse, etc)

mds1 (approved):

components/TokenSelectionModal.tsx
alexkeating: It was mainly testing and specifically the use cases you mentioned above, I haven't used fuse before, but it seemed like the best library for the bundle hit. .1 is probably better because it is more restrictive after taking another look at the docs.

https://fusejs.io/concepts/scoring-theory.html#fuzziness-score

components/TokenInput.tsx
alexkeating: Seems to still pop up errors when the plugin is added. I will look at other configs

components/TokenSelectionModal.tsx
mds1: I've used fuse in the past, it is a nice lib and very configurable. But yea 0.1 or 0.2 both work for me, we can always change it easily in the future

components/TokenInput.tsx
alexkeating: Wasn't able to fix quickly will add back the comments and made a ticket

wildmolasses: sounds good!

Issue #18, opened 2023-01-04
UI: Buiild popup error modal


wildmolasses:
Let's build a popup error modal that can be triggered by other components, for example #14

Example usage would be great!

PR #68, opened 2023-01-26
Add polygon back and removed gasEstimation

### LINKED ISSUE #60: Get gas right on polygon and reenable
Let's see what conclusion results from sdk suggestions. If gas estimation is no longer a prereq, then great; otherwise we can use something like polygon gas station to calculate gas.

c.f. @mds1 element chat:

> https://github.com/ethers-io/ethers.js/blob/bf0b468490cb293cd916e4fff06e0909273719e6/packages/abstract-provider/src.ts/index.ts#L252
> so yea, 1.5gwei is just hardcoded by ethers
> but we can just copy their logic
> ```
> maxPriorityFeePerGas = BigNumber.from(30 gwei);
> maxFeePerGas = block.baseFeePerGas.mul(2).add(maxPriorityFeePerGas);
> ```
> related open issue: https://github.com/ethers-io/ethers.js/issues/3370
> another one https://github.com/ethers-io/ethers.js/issues/2828
> seems people use https://wiki.polygon.technology/docs/develop/tools/polygon-gas-station/ manually for polygon
> nice looking implementation here: https://github.com/ethers-io/ethers.js/issues/2828#issuecomment-1200280374
> good explanation from ricmoo here: https://github.com/ethers-io/ethers.js/issues/2828#issuecomment-1283014250

alexkeating:
#### Description
- I added back polygon and removed the gas estimation support. Also, added some error checks to prevent getting in weird error states.

Closes #60


hooks/useRailgunTx.tsx
```diff
@@ -140,24 +86,23 @@ const useRailgunTx = () => {
 
     // Public wallet to shield from.
     transactionRequest.from = address;
+    transactionRequest.gasPrice = undefined;
+    transactionRequest.gasLimit = undefined;
```
alexkeating, line 90: gasPrice and gasLimit are set to a BigNumber of 0 which was preventing sendTransaction from overriding these values. Probably good feedback for the Railgun team which I can send over tomorrow.

mds1: IIUC the default `transactionRequest` we get back with the new SDK version has them set to zero, so wallets don't do the estimation, and we fix that be re-setting them to `undefined` here? If so, makes sense. Let's leave this unresolved until we get that feedback over to them so we don't forget

components/ReviewTransactionModal.tsx
```diff
@@ -37,7 +37,7 @@ const ReviewTransactionModal = ({
   tokenSymbol,
   tokenName,
 }: ReviewTransactionModalProps) => {
-  const { txNotify } = useNotifications();
+  const { txNotify, notifyUser } = useNotifications();
```
mds1, line 40: _(arbitrary comment location to start a thread)_

Getting this error in the console, that token address is [MATIC](https://polygonscan.com/address/0x0000000000000000000000000000000000001010), i.e. the native token (which exists at that address since ETH is not the native token IIRC)

image

pages/_app.tsx
```diff
@@ -31,7 +31,8 @@ const { chains, provider, webSocketProvider } = configureChains(
       ...bsc,
       iconUrl: bscIcon,
     },
-    // polygon,
+    polygon,
+    arbitrum,
```
mds1, line 35: Arbitrum still isn't supported by railgun yet, right? Just tried using arbitrum and got a bunch of errors, but I think they'll fix themselves once we upgrade to an SDK version that supports it

mds1 (commented):
Polygon transactions working great!

pages/_app.tsx
alexkeating: Yes, I checked their SDK and arbitrum is still commented out. So, I think they are running a little behind on adding support to the SDK because I think their target was the 24th.

hooks/useRailgunTx.tsx
alexkeating: Yes, exactly

hooks/useRailgunTx.tsx
```diff
@@ -44,40 +33,17 @@ const useRailgunTx = () => {
...
-    throw new Error("Unsupported gas type for chain");
+    try {
+      const resp =
+        args.tokenAddress === ethAddress
+          ? await shieldBaseToken(args)
+          : await shieldToken(args);
+      setIsShielding(false);
+      return resp;
+    } catch (e) {
+      console.error(e);
```
wildmolasses, line 44:
```suggestion
      throw e;
```

in fact, we could remove the trycatch entirely, and let the error automatically bubble up to caller

components/ReviewTransactionModal.tsx
```diff
@@ -52,7 +52,14 @@ const ReviewTransactionModal = ({
       tokenDecimals,
       recipient,
     });
-    txNotify(tx!.hash);
+    if (tx) {
+      txNotify(tx.hash);
+    } else {
+      notifyUser({
```
wildmolasses, line 58: feels like toast might be extended past its ux boundary. think we could come back around for a refactor via #18. Also, in tandem with my other comment, `shield` should throw if we don't get a tx, and `doSubmit` should catch it and display the error in, say, the error modal.

wildmolasses (commented):
awesome work! two related comments about error handling

hooks/useRailgunTx.tsx
alexkeating: The main reason I added the try catch was the form was stuck in the loading state because isShielding wasn't set back to false on errors. One other thing, I can trigger an error by cancelling the transaction in my wallet, and that was the main reason I moved from throwing to logging. I am happy to add back the throw, but we probably want to handle a transaction cancellation more gracefully.

components/ReviewTransactionModal.tsx
alexkeating: Thanks! Good catch

pages/_app.tsx
alexkeating: I will remove it from the wallet list

hooks/useRailgunTx.tsx
mds1: Saw you sent it, so marking resolved

components/ReviewTransactionModal.tsx
alexkeating: Fixed this by removing native token from the token list

mds1 (approved):
formatting check in CI is failing, otherwise LGTM

PR #69, opened 2023-01-26
Fix prettier config and update formatting

alexkeating:
#### Description
- Fix prettier config and make sure check command fails


.prettierignore
```diff
@@ -1 +1,2 @@
 .next
+*.yaml
```
mds1, line 2: why not format yaml files anyway?

alexkeating: Good question, bad assumption on my part. Thought prettier didn't support YAML

mds1 (approved):

PR #70, opened 2023-01-27
Feature/support custom tokens

### LINKED ISSUE #13: Token dropdown: add custom token with warning

### LINKED ISSUE #14: Token dropdown: prevent user from entering blacklisted token
from reqs:
> Cannot select something on token blacklist.

Official blacklist to come from railgun, but meanwhile we could create our own minimal blacklist.

Error should be displayed in select modal.

### LINKED ISSUE #18: UI: Buiild popup error modal

alexkeating:
#### Description

- If a user inputs an address in the search and no results are returned then the result will be added as a token item. If the token isn't blacklisted the user should see a warning. If it is blacklisted they will see a different warning and be unable to select it.

[Screencast from 2023-01-27 00-17-16.webm](https://user-images.githubusercontent.com/12705146/215015440-54ac4791-c911-4203-a9dc-7120ea4a5a87.webm)

Closes #13 
Closes #14 
Closes #18


components/AlertModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+type AlertModalProps = {
+  isOpen: boolean;
+  onClose: () => void;
+  onClick: () => void;
+  address: Address;
+  description: string;
+  children?: React.ReactNode;
+};
+
+const AlertModal = (props: AlertModalProps) => {
```
alexkeating, line 24: Should this be changed to ErrorModal? My thought was we could support different levels of alerts like info or warning.

utils/networks.ts
```diff
@@ -69,6 +71,9 @@ export const networks = {
         },
       ],
     },
+    tokenBlacklist: new Map([
+      ['0xc94dd466416A7dFE166aB2cF916D3875C049EBB7', '0xc94dd466416A7dFE166aB2cF916D3875C049EBB7'],
+    ]),
```
alexkeating, line 76: This is temporary and we can handle differently once we have the token blacklist.

alexkeating: #71

mds1: You mean this choice of two tokens is temporary for testing right? If so maybe just add a `// TODO these are temporary, see https://github.com/ScopeLift/railgun-shielder/issues/71" style comment

components/AlertModal.tsx
mds1: I'm indifferent here, will defer to you and @wildmolasses

components/AlertModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <Modal blockScrollOnMount={false} isOpen={props.isOpen} onClose={props.onClose} isCentered>
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader></ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>
+          <Flex direction="column">
+            <Flex justify="center" mb="1rem" color="red.500">
+              <Flex bg="red.100" px=".50rem" py=".25rem" alignItems="center" borderRadius=".75rem">
```
mds1, line 34: I think we should use a yellow color instead here since this is a warning, and red usually means error

components/AlertModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+            {props.children}
+          </Flex>
+        </ModalBody>
+        <ModalFooter>
+          <Button colorScheme="blue" mr={3} onClick={props.onClick}>
+            I understand
+          </Button>
+          <Button variant="ghost" onClick={props.onClose}>
+            Close
+          </Button>
```
mds1, line 53: Seems regardless of which button you click it does the same thing? Since we already have the X icon in the top right, maybe we remove "Close" and have a single "I understand" button

components/TokenSelectionModal.tsx
```diff
@@ -63,18 +79,149 @@ const TokenSelectionItem = ({ token, onClick }: TokenSelectionItemProps) => {
...
+                  textOverflow="ellipsis"
+                  overflow="hidden"
+                  whiteSpace="nowrap"
+                >
+                  {tokenLink}
+                </Text>
+              </Link>
+              <Copy text={tokenLink}>
+                <CopyIcon cursor="pointer" />
+              </Copy>
```
mds1, line 203: We could probably get rid of the copy functionality/component to simplify things and just keep the "open in Etherscan" part. Since there isn't much of a use case for copying the Etherscan link and it's not a patter I've really seen before

utils/networks.ts
alexkeating: Yes, exactly will do

components/AlertModal.tsx
```diff
@@ -0,0 +1,60 @@
+import { Button } from '@chakra-ui/button';
```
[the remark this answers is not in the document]
alexkeating: #72

utils/networks.ts
alexkeating: And also, this is a single token

components/AlertModal.tsx
wildmolasses: sounds good for this PR, we can refactor later

mds1 (approved):

components/AlertModal.tsx
alexkeating: I ended up renaming to WarningModal yesterday

PR #73, opened 2023-01-27
Update SDK to remove overriding gas values

alexkeating:
#### Description

* Incorporate SDK changes for transaction request


package.json
```diff
@@ -15,8 +15,8 @@
     "@chakra-ui/react": "^2.4.6",
     "@emotion/react": "^11",
     "@emotion/styled": "^11",
-    "@railgun-community/quickstart": "3.9.2",
-    "@railgun-community/shared-models": "3.7.5",
+    "@railgun-community/quickstart": "3.8.12",
```
mds1, line 18: Odd that we had to downgrade, wonder how they are managing the versioning

mds1 (approved):
Just tested a shield on polygon and it worked ✅

package.json
alexkeating: Yes, I think they messed up their versioning, there are a couple 3.9 and one 3.10, but this is the latest

PR #74, opened 2023-01-27
Add send page with dynamic Address

### LINKED ISSUE #21: Resolve `/send/0zk1...f` URLs so that they prefill the recipient address
From reqs:

Should work appropriately for both railgun addresses and unstoppable domain. if the address/UD domain is malformed, use the error modal from #18 to alert.

alexkeating:
#### Description

* Fills in address field our current test address locally would now be http://localhost:3000/send/0zk1qyn0qa5rgk7z2l8wyncpynmydgj7ucrrcczhl8k27q2rw5ldvv2qrrv7j6fe3z53ll5j4fjs9j5cmq7mxsaulah7ykk6jwqna3nwvxudp5w6fwyg8cgwkwwv3g4.

Closes #21

### DIFF
--- a/utils/address.ts
+++ b/utils/address.ts
@@ -1,6 +1,5 @@
 import { assertValidRailgunAddress } from '@railgun-community/quickstart';
 
 export const shortenAddress = (address: string) => {
-  assertValidRailgunAddress(address); // Will throw error on incorrect value
   return `${address.slice(0, 7)}...${address.slice(-4)}`;
 };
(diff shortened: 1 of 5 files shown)


utils/address.ts
```diff
@@ -1,6 +1,5 @@
 import { assertValidRailgunAddress } from '@railgun-community/quickstart';
 
 export const shortenAddress = (address: string) => {
-  assertValidRailgunAddress(address); // Will throw error on incorrect value
```
alexkeating, line 4: This was causing an error in the form when typing an address. This validation will be handled by the form validation.

alexkeating:
Closing this in favor of #75

Issue #33, opened 2023-01-11
Refine styling in the app


alexkeating:
In order to move quickly we are moving forward with Chakra's default styles. Go through and make sure the styling matches with the theme we choose and provides a good UX experience. So, evaluate color, font, anchor tag stylings etc...

wildmolasses:
style refined!

PR #75, opened 2023-01-30
Add form validation, clearing on submit and support send page

### LINKED ISSUE #21: Resolve `/send/0zk1...f` URLs so that they prefill the recipient address

### LINKED ISSUE #66: Clear form state and validate form
The form state does not clear on network change and we are missing validation for all the form fields

alexkeating:
#### Description

- Added form validation on amount and address
- Clear fields on submit
- Change selected token on chain change
- Support specifying send in the url (description in #74 )

Closes #66 
Closes #21 

[Screencast from 2023-01-29 21-42-35.webm](https://user-images.githubusercontent.com/12705146/215376638-bf056c73-b572-4033-8331-e05e1746ce0f.webm)


utils/networks.ts
```diff
@@ -135,3 +138,15 @@ export const getEtherscanUrl = (txHashOrAddress: string, chainId: number) => {
...
+
+export const buildBaseToken = (baseToken: BaseToken, chainId: number) => {
+  return {
+    chainId,
+    symbol: baseToken.symbol,
+    address: ethAddress,
+    decimals: 18,
+    name: baseToken.name,
+    logoURI: baseToken.logoURI,
+    balance: BigNumber.from(0), // TODO: Add balance fetch
```
alexkeating, line 150: It is okay if we default this to 0 right? It seems to be overridden in the hook, but want to call it out here.

wildmolasses: here's the part hook where balance is set:

https://github.com/ScopeLift/railgun-shielder/blob/main/hooks/useTokenBalances.tsx#L42-L46

Yeah, I think 0 makes sense as a default. Note that currently, i'm not sure if our useTokenBalances "native" balance calculation is appropriate for all networks. I'll make an issue to track.

components/TxForm.tsx
```diff
@@ -37,94 +56,138 @@ export const TxForm = () => {
...
+                onChange: (e) => {
+                  if (e.target.value && !isNaN(e.target.value)) {
+                    setTokenAmount(e.target.value);
+                  }
+                },
+                validate: (value) => {
+                  try {
+                    return (
+                      Boolean(
+                        parseUnits(value || '0', selectedToken?.decimals).gte(BigNumber.from('0'))
```
mds1, line 144: Is there a reason to allow shielding with a value of 0?
```suggestion
                        parseUnits(value || '0', selectedToken?.decimals).gt(BigNumber.from('0'))
```

components/TxForm.tsx
```diff
@@ -37,94 +56,138 @@ export const TxForm = () => {
...
+    setTokenAmount(values.amount);
+    onReviewOpen();
+  });
+
+  const updateOnNetworkChange = useCallback(
+    (net: GetNetworkResult) => {
+      if (net && net?.chain) {
+        const chain = networks[net.chain.id];
+        const token = buildBaseToken(chain.baseToken, net.chain.id);
+        setSelectedToken(token);
```
mds1, line 77: How come on network change we change the token, but nothing else? We could keep the token selection if the token also exists on the new chain. However if this is a nice to have, not a requirement, we can create an issue and track separately

components/TokenInput.tsx
```diff
@@ -1,50 +1,52 @@
...
+      onOpen: onTokenSelectionOpen,
+      onClose: onTokenSelectionClose,
+    } = useDisclosure();
+    const localOnSelect = (token: TokenListContextItem) => {
+      onSelect(token);
+      onTokenSelectionClose();
+    };
+    return (
+      <>
+        <InputGroup cursor="pointer" onClick={() => onTokenSelectionOpen()}>
```
wildmolasses, line 26: nit
```suggestion
        <InputGroup cursor="pointer" onClick={onTokenSelectionOpen}>
```

components/TokenInput.tsx
```diff
@@ -1,50 +1,52 @@
...
+          <Input
```
wildmolasses, line 27: think the display might be off a little bit: perhaps we can display the token icon and/or choose a different alignment for the token name text. should I create a new issue?

components/TokenInput.tsx
```diff
@@ -1,50 +1,52 @@
...
-      />
-    </>
-  );
-};
+      </>
+    );
+  }
+);
+
+TokenInput.displayName = 'TokenInput';
```
wildmolasses, line 50: what's this for?

components/TxForm.tsx
```diff
@@ -1,35 +1,54 @@
...
+    setValue,
+    reset,
+    formState: { errors },
+  } = useForm<TxFormValues>({
+    defaultValues: {
+      token: network.baseToken.name,
+      recipient: recipientAddress,
+    },
+  });
   const { isOpen: isReviewOpen, onOpen: onReviewOpen, onClose: onReviewClose } = useDisclosure();
```
wildmolasses, line 47: nit (due to time, would probably ignore)

```suggestion
  const { isOpen: isReviewOpen, onOpen: openReview, onClose: closeReview } = useDisclosure();
```

components/TxForm.tsx
wildmolasses: i'm a little confused here for different reasons. what does `updateNetworkChange` achieve?

is this complexity due to https://github.com/ScopeLift/railgun-shielder/pull/65#discussion_r1086971131 or something else?

components/TxForm.tsx
```diff
@@ -38,95 +57,138 @@ export const TxForm = () => {
...
-            type="number"
             variant="outline"
             size="lg"
             pr="4.5rem"
-            height="100%"
-            value={tokenAmount}
-            onChange={(e) => setTokenAmount(e.target.value)}
+            height="4rem"
+            mb=".25rem"
+            {...register('recipient', {
```
wildmolasses, line 107: looks great! nit: In this PR we may have hit the threshold for component logic in the jsx markup. let's call a refactor here "nice to have" if needed at all. not merge-blocking!

components/ReviewTransactionModal.tsx
```diff
@@ -12,43 +12,39 @@ import {
...
   onClose,
-  tokenAddress,
   recipient,
-  tokenAmount,
-  tokenDecimals,
-  tokenSymbol,
-  tokenName,
+  amount,
+  token,
+  onClick,
```
wildmolasses, line 35: what kind of click are we talking about here? might prefer a more descriptive name (onSubmitClick?)

wildmolasses (commented):
some comments, nothing big, thanks for driving this forward @alexkeating !!

components/TxForm.tsx
alexkeating: Good question, I don't think so, we can surface a different error message for a 0

components/TxForm.tsx
alexkeating: Yes, it is due to that comment. And this resets the the selected token to the native token on network change because it is the simplest way to guarantee a chain compatible token. We could add additional logic if we want to see if we can use the same token on the previous chain like Matt suggested, but I would say this is a nice to have and can be handled in a separate ticket.

components/TokenInput.tsx
```diff
@@ -1,50 +1,52 @@
...
-    <>
-      <InputGroup cursor="pointer" onClick={() => onTokenSelectionOpen()}>
-        <Input
-          type="button"
-          pr="4.5rem"
-          size="lg"
-          height="4rem"
-          value={token?.name || tokenList[0]?.name || ''} // Default to native which is assumed at the top of the token list
-          mb=".75rem"
+const TokenInput = React.forwardRef(
```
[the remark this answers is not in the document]
alexkeating: I think that is the right way to integrate a custom component in react hook form. The other way is via a controlled component, but I think we should prefer the forward ref.

https://react-hook-form.com/get-started#Integratinganexistingform

https://react-hook-form.com/get-started#IntegratingControlledInputs

components/TokenInput.tsx
alexkeating: Yea, create a new ticket for that probably. I realigned the carat after looking at this comment.

components/TokenInput.tsx
alexkeating: This is to appease the linter. I think the forward ref makes the display an empty string

components/ReviewTransactionModal.tsx
alexkeating: Good call will change

components/TokenInput.tsx
alexkeating: #33 Should be captured here

components/TxForm.tsx
wildmolasses: part of me thinks we should just let "no token" be selected after all, as the alternative seems a bit bloaty, and to what gain? will defer to you @alexkeating, what do you think?

components/TokenInput.tsx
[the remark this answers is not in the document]
wildmolasses: thanks!

components/TxForm.tsx
alexkeating: My vote is to keep it for now if the following is right. I think we will need some network switching logic regardless if it is filled out or not. 

1, If we have an empty value and a user selects a token and then they switch networks they could have an invalid token value unless we clear on network switch. And if we clear, we will need something like this to clear the form state on network change if that is the behavior we want instead.

wildmolasses: sounds good, but couldn't it be a simple `useEffect` hook where the callback sets the token state to `undefined`? or, is it more complicated now that we're using the ref with the react-hook-form? Anyway, I think i'll resolve this in favor of moving things along 👍

wildmolasses (approved):
lgtm!!

PR #78, opened 2023-01-30
feat: load shield fees instead of hardcoding it

### LINKED ISSUE #61: Change fee amount per network change
Currently the fee amount used in the review modal is hard coded, and it should instead change based on the connected network.

mds1:
Closes #61 

React feedback/criticisms welcome 🙂

### DIFF
--- a/components/ReviewTransactionModal.tsx
+++ b/components/ReviewTransactionModal.tsx
@@ -38,13 +38,14 @@ const ReviewTransactionModal = ({
 }: ReviewTransactionModalProps) => {
   const { txNotify, notifyUser } = useNotifications();
   const { shield, isShielding } = useRailgunTx();
-  const { shieldFees } = useRailgunProvider();
+  const { shieldingFees } = useRailgunProvider();
   const { chain } = useNetwork();
   const tokenAmount = amount;
   const tokenDecimals = token?.decimals;
-  const bigNumberAmount = parseUnits(tokenAmount || '0', tokenDecimals);
-  const shieldFee = shieldFees[chain?.id || 1];
-  const feeAmount = parseUnits(tokenAmount || '0', tokenDecimals)
+
+  const bigNumberAmount = parseUnits(tokenAmount! || '0', tokenDecimals);
+  const shieldFee = shieldingFees[chain?.id || 1];
+  const feeAmount = parseUnits(tokenAmount! || '0', tokenDecimals)
     .mul(shieldFee)
     .div(10000);
 
--- a/components/ReviewTransactionModal.tsx
+++ b/components/ReviewTransactionModal.tsx
@@ -12,8 +12,10 @@ import {
 import { Tooltip } from '@chakra-ui/tooltip';
 import { ethers } from 'ethers';
 import { parseUnits } from 'ethers/lib/utils.js';
+import { useNetwork } from 'wagmi';
 import { TokenListContextItem } from '@/contexts/TokenContext';
 import useNotifications from '@/hooks/useNotifications';
+import { useRailgunProvider } from '@/hooks/useRailgunProvider';
 import useRailgunTx from '@/hooks/useRailgunTx';
 import { shortenAddress } from '@/utils/address';
 
@@ -36,8 +38,15 @@ const ReviewTransactionModal = ({
 }: ReviewTransactionModalProps) => {
   const { txNotify, notifyUser } = useNotifications();
   const { shield, isShielding } = useRailgunTx();
-  const bigNumberAmount = parseUnits(amount! || '0', token?.decimals);
-  const feeAmount = parseUnits(amount! || '0', token?.decimals).div('400');
+  const { shieldFees } = useRailgunProvider();
+  const { chain } = useNetwork();
+  const tokenAmount = amount;
+  const tokenDecimals = token?.decimals;
+  const bigNumberAmount = parseUnits(tokenAmount || '0', tokenDecimals);
+  const shieldFee = shieldFees[chain?.id || 1];
+  const feeAmount = parseUnits(tokenAmount || '0', tokenDecimals)
+    .mul(shieldFee)
+    .div(10000);
 
   const doSubmit: React.FormEventHandler = async () => {
     if (!token.address || !amount || !token?.decimals || !recipient) throw new Error('bad form');
--- a/hooks/useRailgunProvider.tsx
+++ b/hooks/useRailgunProvider.tsx
@@ -10,16 +10,16 @@ interface ShieldFee {
   [chainId: number]: BigNumber;
 }
 
-const initialShieldFees: ShieldFee = {};
+const fallbackShieldingFees: ShieldFee = {};
 Object.keys(networks).forEach((chainId) => {
   // Current fees are 0.25% everywhere, so we initialize with that as a default -- If something goes
   // wrong fetching fees, better to show a higher fee than a zero default fee.
-  initialShieldFees[Number(chainId)] = BigNumber.from('25');
+  fallbackShieldingFees[Number(chainId)] = BigNumber.from('25');
 });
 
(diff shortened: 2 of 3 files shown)


hooks/useRailgunProvider.tsx
```diff
@@ -1,19 +1,43 @@
...
+
+      // Set the shield fees for each network. For convenience, we just update the
+      // `initialShieldFees` object then set it.
+      res.forEach((r) => {
+        const newFee = r.providerInfo.feesSerialized?.shield;
+        initialShieldFees[r.chainId] = BigNumber.from(newFee || initialShieldFees[r.chainId]);
+      });
+      setShieldFees(initialShieldFees);
+
+      // Provider is done loading.
```
alexkeating, line 40: The one improvement we could make here is to decouple this logic from loading the providers, mainly because anywhere this hook is called the providers will be reloaded. The cleanest way to do this may be to some of this to context, but definitely a future improvement.

alexkeating (approved):

hooks/useRailgunProvider.tsx
mds1: Yea, I was also thinking of wrapping it within `useSWR` but wasn't sure if it's bad to use `useSWR` within a `useEffect` hook?

mds1: Ah just tried, doesn't really fit here since we have the `setProviderLoaded` side effect. Looks like it runs 4 times and then never runs again (even when switching networks), so this probably ok for now

mds1: Now tracking in https://github.com/ScopeLift/railgun-shielder/issues/80

alexkeating: I don't think we should call hooks inside useEffect because it breaks the [rules of hooks](https://reactjs.org/docs/hooks-rules.html).  We could move load providers to the top level and wrap in an SWR hook and I guess it would be cached if it is called multiple times. For network switching did you add the chain id as a value for the cache key? My guess is that was why it didn't update.

mds1: > I don't think we should call hooks inside useEffect because it breaks the [rules of hooks](https://reactjs.org/docs/hooks-rules.html).

Right that doesn't work. To clarify I tried removing `useEffect` and putting the side effect within the `useSWR` fetcher. Sorry, I wasn't clear on that

> For network switching did you add the chain id as a value for the cache key? My guess is that was why it didn't update.

What didn't update on network switching? To clarify, right now `loadProviders` fetches the info for all supported chains, so it only _needs_ to run once (or poll if we think the fee could change during the time a user is using the app, but not worried about that for now). What I meant to convey is that right now, the current implementation of the PR, has the current hook run 4 times (some re-renders occur after initial load I guess), and that's it. So since it's not constantly running it seems ok for now even if not technically optimal

My message was very unclear, sorry 😅

alexkeating: Ah okay, thank you for clarifying!

hooks/useRailgunProvider.tsx
```diff
@@ -1,19 +1,43 @@
...
 
   useEffect(() => {
     const fn = async () => {
       setProviderLoaded(false);
-      await loadProviders();
+      const res = await loadProviders();
+
+      // Set the shield fees for each network. For convenience, we just update the
+      // `initialShieldFees` object then set it.
+      res.forEach((r) => {
```
wildmolasses, line 34: style nit: don't love mutating the `initialShieldFees` object.

```js
const shieldFeesFromNetwork = res.reduce((acc, response) => {
        const newFee = response.providerInfo.feesSerialized?.shield; // Personally think we should error loudly here if `feesSerialized` not available
        return {
           ...acc,
           [response.chainId]: BigNumber.from(newFee || initialShieldFees[response.chainId])
        }
}, {})

setShieldFees(shieldFeesFromNetwork)
```

hooks/useRailgunProvider.tsx
```diff
@@ -23,5 +47,5 @@ export const useRailgunProvider = () => {
     setProviderForNetwork(network.railgunNetworkName, provider);
   }, [provider, network.railgunNetworkName]);
 
-  return { isProviderLoaded };
+  return { isProviderLoaded, shieldFees };
```
wildmolasses, line 50: nit: some slight ambiguity since "shield" is a verb everywhere else

```suggestion
  return { isProviderLoaded, shieldingFees };
```

components/ReviewTransactionModal.tsx
```diff
@@ -39,8 +41,14 @@ const ReviewTransactionModal = ({
 }: ReviewTransactionModalProps) => {
   const { txNotify, notifyUser } = useNotifications();
   const { shield, isShielding } = useRailgunTx();
+  const { shieldFees } = useRailgunProvider();
+  const { chain } = useNetwork();
+
   const bigNumberAmount = parseUnits(tokenAmount! || '0', tokenDecimals);
-  const feeAmount = parseUnits(tokenAmount! || '0', tokenDecimals).div('400');
+  const shieldFee = shieldFees[chain?.id || 1];
+  const feeAmount = parseUnits(tokenAmount! || '0', tokenDecimals)
```
wildmolasses, line 49: why do we divide by 10000? is that the unit `shieldFee` is denominated in?

wildmolasses (commented):
thanks @mds1 , a couple comments

hooks/useRailgunProvider.tsx
```diff
@@ -1,19 +1,43 @@
...
 
+// Fee is in bips, e.g. a value of 25 is a 0.25% fee.
+interface ShieldFee {
+  [chainId: number]: BigNumber;
+}
+
+const initialShieldFees: ShieldFee = {};
+Object.keys(networks).forEach((chainId) => {
+  // Current fees are 0.25% everywhere, so we initialize with that as a default -- If something goes
+  // wrong fetching fees, better to show a higher fee than a zero default fee.
```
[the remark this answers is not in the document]
mds1: If something goes wrong fetching fees, i.e. the `loadProvider` call, the app can't work anyway though IIUC?

hooks/useRailgunProvider.tsx
mds1: refactored in 95365db

> // Personally think we should error loudly here if `feesSerialized` not available

If `loadProvider` returns successfully this will never be not available (just asked railgun) so the right spot to handle this is in `railgun.ts`. We weren't handling the error case there originally, so worth pulling out into a separate issue? (related to https://github.com/ScopeLift/railgun-shielder/pull/78#discussion_r1092323102)

hooks/useRailgunProvider.tsx
mds1: Renamed in f7eb5d2

components/ReviewTransactionModal.tsx
mds1: Yea, because the units are bips https://github.com/ScopeLift/railgun-shielder/blob/166be6c4270df9b4c85724df3c128667a350c934/hooks/useRailgunProvider.tsx#L8

wildmolasses: thanks!

hooks/useRailgunProvider.tsx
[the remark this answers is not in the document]
wildmolasses: in that case, the comment is a little misleading, because "if something goes wrong fetching fees," the app won't work (and won't show a higher fee than a zero default fee), right? might suggest truncating to:

```
// Current fees are 0.25% everywhere, so we initialize with that as a default 
```

hooks/useRailgunProvider.tsx
wildmolasses: tracking in #85, thanks!

Issue #24, opened 2023-01-04
URL og metadata


wildmolasses:
from reqs:

> Metadata for URL links on Twitter, Telegram, Discord

mds1:
As part of this issue let's also fix the favicon 
<img width="441" alt="image" src="https://user-images.githubusercontent.com/17163988/215100691-40070b50-cb87-4155-a6e4-131eb69fe61b.png">

alexkeating:
Do we have logos yet? And will the social information be for Railgun or specific to the shielder?

apbendi:
Update title to say "Token Shielder" instead of "Railgun Shielder" @mds1

PR #83, opened 2023-02-01
feat: Open Graph tags

mds1:
Closes https://github.com/ScopeLift/railgun-shielder/issues/24

Previews from [Sizzy](https://sizzy.co/):

image


pages/_app.tsx
```diff
@@ -61,9 +64,19 @@ function MyApp({ Component, pageProps }: AppProps) {
...
+        <title>{APP_TITLE}</title>
+        <meta name="description" content={APP_DESCRIPTION} />
+        <link rel="icon" href="protect-icon.svg" />
+
+        <meta property="og:title" content={APP_TITLE} />
+        <meta property="og:type" content="website" />
+        <meta property="og:url" content="https://tokenshielder.com" />
+        <meta
+          property="og:image"
+          content="https://raw.githubusercontent.com/ScopeLift/railgun-shielder/9d8b5dda5826c8121a9934da484a29645de161e6/public/home.png"
```
mds1, line 76: You can't point to local files here so I pushed a commit that uploaded the file, got its permalink, then pushed another commit. More info here: https://stackoverflow.com/a/74917023

Another option is we can convert the png file to base64 and inline it here, or the alternatives discussed here, but this approach seems fine. Also note that the image is a bit chunky at 1.3 MB but it's not part of the actual bundle so that seems ok?

alexkeating (approved):
This looks good to me, the image content seems fine for what we need

wildmolasses (approved):
lgtm, re: png size we can optimize later if necessary!

PR #84, opened 2023-02-01
Add copy send link button

### LINKED ISSUE #20: Create "URL link generator" component
From reqs:

Let's create a small component that takes an input value and generates a send link for the user.

alexkeating:
#### Description

- Add button with outline style that adds the link to the clipboard.
- The form pr should probably get in before this one

Closes #20

### DIFF
--- a/components/TxForm.tsx
+++ b/components/TxForm.tsx
@@ -2,8 +2,9 @@ import React, { useCallback, useEffect, useState } from 'react';
 import { useForm } from 'react-hook-form';
 import { Button } from '@chakra-ui/button';
 import { FormControl, FormErrorMessage, FormLabel } from '@chakra-ui/form-control';
+import { CopyIcon } from '@chakra-ui/icons';
 import { Input, InputGroup } from '@chakra-ui/input';
-import { Box } from '@chakra-ui/layout';
+import { Box, Flex, Text } from '@chakra-ui/layout';
 import { useDisclosure } from '@chakra-ui/react';
 import { getRailgunSmartWalletContractForNetwork } from '@railgun-community/quickstart';
 import { validateRailgunAddress } from '@railgun-community/quickstart';
@@ -64,6 +65,13 @@ export const TxForm = ({ recipientAddress }: { recipientAddress?: string }) => {
   const needsApproval =
     selectedToken?.address !== ethAddress &&
     ethers.utils.parseUnits(tokenAmount || '0', selectedToken?.decimals).gt(tokenAllowance);
+  const onCopy = () => {
+    navigator.clipboard.writeText(`${window.location.host}/send?address=${recipient}`);
+    notifyUser({
+      alertType: 'success',
+      message: 'Shield link copied to clipboard',
+    });
+  };
   const onSubmit = handleSubmit(async (values) => {
     setRecipient(values.recipient);
     setTokenAmount(values.amount);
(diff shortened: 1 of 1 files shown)


wildmolasses (commented):
Functionally spot on. UX-wise I'm wondering where else we can stick this button, as it currently competes w/ the primary functionality. How about closer to the recipient field, like in the image here? Maybe it only appears when there's text in the field (or conditionally on address validity). Deferring to you on this.

image

mds1:
@wildmolasses for context the current design is based on Umbra's, shown below. I think the combination of:
1. Making the shield button blue (same blue as the "connect wallet" button)
2. Removing the outline from the button
3. Making the blue instead of black 

would make it compete less with the primary shield button

image

wildmolasses:
> @wildmolasses for context the current design is based on Umbra's, shown below. I think the combination of:
> 
> 1. Making the shield button blue (same blue as the "connect wallet" button)
> 2. Removing the outline from the button
> 3. Making the blue instead of black
> 
> would make it compete less with the primary shield button

i generally agree, although i still think the "copy link" belongs closer to the recipient address, as it's that specific field that the action bears relevance to, and not so much the completed form. that said, i'll defer to implementer.

(also, sorry, this is the last time i use unthreaded comments!)

components/TxForm.tsx
```diff
@@ -127,6 +135,9 @@ export const TxForm = () => {
           Shield
         </Button>
       )}
+      <Button variant="outline" size="lg" mt="1rem" width="100%" onClick={onCopy}>
+        Copy Send Link
```
alexkeating, line 139: (arbitrary comment) @mds1 @wildmolasses 

I think moving the functionality to where Ed wants makes sense, but instead of text can we use a share logo?

wildmolasses: is there an icon you have in mind? do you think the functionality would be apparent from the icon, and if not could we pair it with text?

mds1: Cool I'm good with @wildmolasses's location. For an icon, usually a copy icon looks something like [this](https://fontawesome.com/icons/copy?s=solid&f=classic) (we can use whatever icon library you guys like).

`Copy Send Link <icon>` seems good, I don't love icons with no text because personally I find it confusing UX, unless it tells you what it does with a tooltip on hover

components/TxForm.tsx
```diff
@@ -47,6 +47,14 @@ export const TxForm = () => {
     tokenAddress !== ethAddress &&
     ethers.utils.parseUnits(tokenAmount || '0', tokenDecimals).gt(tokenAllowance);
 
+  const onCopy = () => {
+    navigator.clipboard.writeText(`${window.location.hostname}/send?address=${recipient}`);
+    notifyUser({
+      alertType: 'success',
+      message: 'Added link to clipboard',
```
mds1, line 54:
```suggestion
      message: 'Shield link copied to clipboard',
```

utils/address.ts
```diff
@@ -1,6 +1,3 @@
-import { assertValidRailgunAddress } from '@railgun-community/quickstart';
-
 export const shortenAddress = (address: string) => {
-  assertValidRailgunAddress(address); // Will throw error on incorrect value
```
mds1, line 4: Why did we get rid of this?

components/TxForm.tsx
alexkeating: I was thinking about this icon, https://fontawesome.com/icons/share-nodes?s=light&f=classic. I am down to keep it with text.

utils/address.ts
alexkeating: The main reason was if it errors it crashes the app, and we use it in the jsx. So, If we leave it this way, we will need to add some logic within the component, and the check isn't necessary because we are using form validation.

alexkeating: It is simpler to remove rather than work around it.

components/TxForm.tsx
```diff
@@ -64,6 +65,13 @@ export const TxForm = ({ recipientAddress }: { recipientAddress?: string }) => {
   const needsApproval =
     selectedToken?.address !== ethAddress &&
     ethers.utils.parseUnits(tokenAmount || '0', selectedToken?.decimals).gt(tokenAllowance);
+  const onCopy = () => {
+    navigator.clipboard.writeText(`${window.location.host}/send?address=${recipient}`);
```
mds1, line 69: Probably should error/not copy if the address is empty or invalid? Feel free to pull this out into a separate issue

wildmolasses: tracking in #87

wildmolasses (approved):
lgtm

PR #88, opened 2023-02-02
Add blacklist with rebase tokens

### LINKED ISSUE #71: Add token blacklist
We currently have a mock token blacklist and we need to add the real ones.

alexkeating:
#### Description

- Disallow rebase tokens from being shielded.

Closes #71


public/blacklist.json
```diff
@@ -0,0 +1,58 @@
+[
+  "0x64aa3364f17a4d01c6f1751fd97c2bd3d7e7f1d5",
```
mds1, line 2: Let's either make this a TS file that exports the array with a comment linking to the source of this array, or document that link in the PR description

components/TokenSelectionModal.tsx
```diff
@@ -136,7 +137,9 @@ const CustomTokenSelectionItem = ({
     token: tokenAddress,
     chainId: chain?.id,
   });
-  const isBlacklisted = network.tokenBlacklist.get(tokenAddress);
+  const isBlacklisted = blacklistJson.find(
```
mds1, line 140: Can we change the blacklist message from the current 

> This token is blacklisted and cannot be shielded!

to something that explains why, such as:

> This token cannot be shielded because it is a rebase token

or 

> This token is a rebase token and therefore cannot be shielded

public/blacklist.json
```diff
@@ -0,0 +1,58 @@
...
+  "0x027dbca046ca156de9622cd1e2d907d375e53aa7",
+  "0xdb021b1b247fe2f1fa57e0a87c748cc1e321f07f",
+  "0xf2f5bf00cd952f3f980a02f5dce278cbff4dae05",
+  "0x4e78011ce80ee02d2c3e649fb657e45898257815",
+  "0x470ebf5f030ed85fc1ed4c2d36b9dd02e77cf1b7",
+  "0x7d1232b90d3f809a54eeaeebc639c62df8a8942f",
+  "0x4a436073552044d5f2f49b176853ad3ad473d9d6",
+  "0xfa1fbb8ef55a4855e5688c0ee13ac3f202486286",
+  "0x5602df4a94eb6c680190accfa2a475621e0ddbdc",
+  "inL8PMVd6iiW3RCBJnr5AsrRN6nqr4BTrcNuQWQSkvY",
```
wildmolasses, line 15: this address looks weird

alexkeating: Yes, agreed should be we just remove it? It is in there list https://github.com/Railgun-Privacy/frontend-react-private/blob/main/src/services/token/exports/rebase_token_address_list.json#L15

mds1: For reference, this is a solana token, so yea seems ok to remove it https://explorer.solana.com/address/inL8PMVd6iiW3RCBJnr5AsrRN6nqr4BTrcNuQWQSkvY

mds1 (approved):

wildmolasses (approved):
lgtm!

PR #89, opened 2023-02-02
Prevent crash on wrong network

alexkeating:
#### Description

- I noticed the app was crashing on an unsupported network because of the networks object. I default us to ethereum to prevent errors and disabled the shield and approve button.

Screenshot from 2023-02-01 23-30-54


components/TxForm.tsx
```diff
@@ -97,6 +97,7 @@ export const TxForm = ({ recipientAddress }: { recipientAddress?: string }) => {
     const unwatch = watchNetwork(updateOnNetworkChange);
     return unwatch;
   }, [updateOnNetworkChange]);
+  console.log(chain);
```
mds1, line 100: Delete?

mds1 (approved):

PR #95, opened 2023-02-03
Prevent duplicate toasts on custom token error

### LINKED ISSUE #90: Bug on custom token that does not resolve to a token
When a user puts in a valid address that is not a token we send a bunch of error toasts. We should probably log rather than pop a toast for these errors.

Screenshot from 2023-02-01 23-41-59

alexkeating:
#### Description
- Prevent duplicate toasts

Closes #90 

Screenshot from 2023-02-03 12-13-07


mds1:
LGTM, how easy is it to make both of those toasts red? What's the difference between the red and orange ones? But they do both seems like errors

mds1 (approved):

PR #99, opened 2023-02-03
Disable shield button when not connected

alexkeating:


components/TxForm.tsx
```diff
@@ -207,7 +208,7 @@ export const TxForm = ({ recipientAddress }: { recipientAddress?: string }) => {
             size="lg"
             mt=".75rem"
             width="100%"
-            isDisabled={chain?.unsupported}
+            isDisabled={!isConnected && chain?.unsupported}
```
mds1, line 211: I think we want an `or` here? So the button is disable if not connected, or if the chain is unsupported
```suggestion
            isDisabled={!isConnected || chain?.unsupported}
```

mds1: Actually I guess both end up behaving the same, since if not connected, the chain is unsupported?

alexkeating: yea, that's what I thought, I think your correction is more correct though

mds1 (approved):

PR #101, opened 2023-02-07
Add back arbitrum

### LINKED ISSUE #64: Add back arbitrum support
* Arbitrum is not yet supported in the SDK. Once it is supported in the SDK we should add it back to the wallet and the networks object

alexkeating:
#### Description
- Add arbitrum to the networks

Closes #64


alexkeating:
@mds1 Updated libraries and added useMemo to make sure loadProviders is only called once

utils/networks.ts
```diff
@@ -99,6 +99,21 @@ export const networks = {
       providers: [{ provider: getRpcUrl(polygon.id), priority: 1, weight: 1 }],
     },
   },
+  [arbitrum.id]: {
+    blockExplorerUrl: 'https://arbiscan.io/',
+    railgunNetworkName: NetworkName.Arbitrum,
+    chainId: arbitrum.id,
+    wethAddress: '0x82aF49447D8a07e3bd95BD0d56f35241523fBab1',
+    baseToken: {
+      symbol: 'AETH',
```
mds1, line 108: Arbitrum docs + popular apps all use `ETH` now (used to be AETH a while ago)
```suggestion
      symbol: 'ETH',
```

package.json
```diff
@@ -15,8 +15,8 @@
     "@chakra-ui/react": "^2.4.6",
     "@emotion/react": "^11",
     "@emotion/styled": "^11",
-    "@railgun-community/quickstart": "3.8.12",
-    "@railgun-community/shared-models": "3.7.6",
+    "@railgun-community/quickstart": "3.11.6",
```
mds1, line 18: _(arbitrary comment location to start a thread)_

If I add `console.log('loading...')` as the first line in `loadProviders` in our `raulgun.ts` method, I still see it being called 4 times

image

alexkeating: That may be an artifact of it running locally, in strict mode. I tested with a console log on the production build and it only showed up once for me. I will push up a console log so we can test.

alexkeating: Ah nvm, I was confused and thought we were targeting the initialize

alexkeating: Seems to be working now, I left a log in to confirm

mds1: I'm seeing it logged twice now, but that's probably fine since it shouldn't be an issue anymore anyway. Let's go with this and see how it goes
image

PR #103, opened 2023-02-19
Improvements around unstoppable domain

### LINKED ISSUE #102: Show copy shield link only with valid address
1. Hide “Copy shield link” action until valid 0zk address / unstoppable domain is filled
2. Validate Recipient Address 0zk address before attempting to submit
3.  Only query unstoppable domains if the address matches a domainish regex, or includes supported TLD
    - check whether they have a regex for us already
4. The query to unstoppable periodically deactivates the submit button (do some comms with railgun)
    - regex check might solve part of this
    - might consider debouncing reqs
    - perhaps no reaasonable way to avoid deactivating submit entirely

alexkeating:
#### Description
- Run validation on change
- Only show `Copy link` when valid address
- Only call unstoppable when address has a valid ending
- Move unstoppable resolution to validation so we don't need to disable submit
- Remove right padding

Closes #102

### DIFF
--- a/components/TxForm.tsx
+++ b/components/TxForm.tsx
@@ -42,9 +42,11 @@ export const TxForm = ({ recipientAddress }: { recipientAddress?: string }) => {
     register,
     setValue,
     reset,
+    clearErrors,
+    setError,
     formState: { errors },
   } = useForm<TxFormValues>({
-    reValidateMode: 'onSubmit',
+    mode: 'onChange',
     defaultValues: {
       token: network.baseToken.name,
       recipient: recipientAddress,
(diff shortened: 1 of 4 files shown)


wildmolasses:
nit: can we get the container to not resize on different error messages? otherwise lgtm.

https://user-images.githubusercontent.com/5423313/220425183-1a7238ec-97ce-4a3d-b708-f4c2ec36ff30.mp4

components/TxForm.tsx
```diff
@@ -123,16 +123,18 @@ export const TxForm = ({ recipientAddress }: { recipientAddress?: string }) => {
           <Flex justify="space-between">
             <FormLabel>Recipient address</FormLabel>
 
-            <Text
```
alexkeating, line 126: (arbitray comment)

- I can't reproduce the width changing on my local or in my browser brave/firefox. But, I am seeing the height shift. Should I make the height larger?

wildmolasses: let's punt on it, we can adjust in another PR if necessary

wildmolasses (approved):

PR #105, opened 2023-02-20
Add placeholder text

### LINKED ISSUE #104: Add placeholder text
- Add placeholder text in Recipient address field, to show example 0zk address
- Add placeholder text in Amount field, to show example amount “0.1”
- Can we add in grey text in the Token search box something like ‘Search for token name or paste address’

alexkeating:
#### Description
- Add placeholder text

Closes #104

### DIFF
--- a/components/TxForm.tsx
+++ b/components/TxForm.tsx
@@ -137,9 +137,9 @@ export const TxForm = ({ recipientAddress }: { recipientAddress?: string }) => {
           <Input
             variant="outline"
             size="lg"
-            pr="4.5rem"
             height="4rem"
             mb=".25rem"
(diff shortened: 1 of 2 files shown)


components/TxForm.tsx
```diff
@@ -137,9 +137,9 @@ export const TxForm = ({ recipientAddress }: { recipientAddress?: string }) => {
           <Input
             variant="outline"
             size="lg"
-            pr="4.5rem"
             height="4rem"
             mb=".25rem"
+            placeholder="0zk1qyn0qa5rgk7z2l8wyncpynmy..."
```
alexkeating, line 142: We can display a whole address if we switch to a textarea, but the form will look a little weirder

wildmolasses (approved):
lgtm

PR #115, opened 2023-02-20
Remove max int allowance

### LINKED ISSUE #112: Specify amount in ERC20 approvals rather than unlimited
Also it does only limit approvals - most people are afraid of those. Possible to make it not unlimited approvals?

alexkeating:
#### Description
- Create allowance for amount

Closes #112


wildmolasses (approved):
works for me, thanks

PR #135, opened 2023-03-31
Handle case where amount has too many decimals

### LINKED ISSUE #124: Tx form: should handle the case where user types in crazy amount
Original feedback:
> Entering more decimals of precision than a token supports causes a crash to an error screen.

https://user-images.githubusercontent.com/5423313/228401523-5cdad152-e968-4944-8ad9-ec7f19e258ff.mp4

alexkeating:
#### Description

- Handle when there are too many decimals

[Screencast from 2023-03-30 23-04-24.webm](https://user-images.githubusercontent.com/12705146/229012486-ddd0c12f-c92f-4d08-975a-069e2d640393.webm)

Closes #124

### DIFF
--- a/utils/token.ts
+++ b/utils/token.ts
@@ -1,0 +1,11 @@
+import { ethers } from 'ethers';
+
+export const ethersParsable = (tokenAmount: string, decimals: number) => {
+  try {
+    ethers.utils.parseUnits(tokenAmount || '0', decimals);
+    return true;
+  } catch (e) {
+    console.error(e);
+    return false;
+  }
+};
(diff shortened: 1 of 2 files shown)


utils/token.ts
```diff
@@ -0,0 +1,11 @@
+import { ethers } from 'ethers';
+
+export const ethersParsable = (tokenAmount: string, decimals: number) => {
```
wildmolasses, line 3: nit: since this returns a bool maybe this name change is clearer

```suggestion
export const isAmountParsable = (tokenAmount: string, decimals: number) => {
```

wildmolasses (approved):
lgtm, small nit

PR #138, opened 2023-04-05
Store custom tokens

### LINKED ISSUE #125: Token selection modal: should save custom token in user's token list
waiting to hear: should we save just for the session, or persist between sessions?

> When you add a custom token, it does not save it in the menu of tokens a user can select

https://user-images.githubusercontent.com/5423313/228402041-4c229d8d-a1a0-4b69-8097-11fabc49bbe3.mp4

alexkeating:
#### Description
- Added hook that stores a custom token locally once a user confirms the token on the confirmation modal

[Screencast from 2023-04-05 15-01-44.webm](https://user-images.githubusercontent.com/12705146/230179417-d9ece575-fd7e-4ce9-9968-f7c75648e9a7.webm)

Closes #125


utils/constants.ts
```diff
@@ -4,6 +4,8 @@ export const ethAddress = getAddress(`0x${'e'.repeat(40)}`);
 export const ipfsDomain = 'https://gateway.ipfs.io/ipfs/';
 export const VALID_AMOUNT_REGEX = /^[\d\\.]+$/;
 
+export const CUSTOM_TOKEN_ARRAY_PATH = 'customTokens';
```
wildmolasses, line 7:
```suggestion
export const CUSTOM_TOKENS_STORAGE_KEY = 'customTokens';
```

hooks/useTokenList.tsx
```diff
@@ -17,5 +19,14 @@ export const useTokenList = () => {
...
   const tokenList = tokenListJson.tokens.filter((token) => token.chainId === chainId);
   const baseToken = buildBaseToken(network.baseToken, chain?.id || 1);
-  return { tokenList: [baseToken, ...tokenList] };
+  const {
+    data: localTokenList,
+    isLoading,
+    error,
+  } = useLocalForageGet<TokenListItem[]>({ itemPath: CUSTOM_TOKEN_ARRAY_PATH });
+  const tokens = [baseToken, ...tokenList];
+  if (!isLoading && !error && localTokenList) {
```
wildmolasses, line 28: nit: might prefer a single return statement (feel free to ignore!)

```ts
const localTokens = localTokenList || [];
return { tokenList: [...tokens, ...localTokens] };
```

wildmolasses (approved):
lgtm, a couple nits

another thing i noticed is that when there's no wallet connected, i got an error when adding a custom token. but that feels more appropriate to log as a separate issue, which i'll do.

PR #140, opened 2023-04-06
Common tokens are sorted to the top of the token list

### LINKED ISSUE #131: Token selection modal: Pin most common tokens
> can we pin maybe the network base token, usdc, dai, usdt to the top?

wildmolasses:
Resolves #131


utils/constants.ts
```diff
@@ -4,6 +4,8 @@ export const ethAddress = getAddress(`0x${'e'.repeat(40)}`);
 export const ipfsDomain = 'https://gateway.ipfs.io/ipfs/';
 export const VALID_AMOUNT_REGEX = /^[\d\\.]+$/;
 
+export const TOKEN_PRIORITY_SORT = ['USDC', 'DAI', 'USDT'];
```
alexkeating, line 7: This may be a deficiency in our token list, but every chain looks good except BNB for me.

Screenshot from 2023-04-06 20-40-10

alexkeating (approved):
Looks good, left one comment about BNB list

utils/constants.ts
wildmolasses: thanks -- indeed, our token list does not include USDC, DAI, or USDT for bsc

